PR #271, opened 2021-11-27
feat: 🎸 sm and lg sizing for radio and checkbox

mario-subo:
Should I be trying to merge to `main`, `alpha` or `beta`? 

Input label
Input label
Input label
Large input

Inspiration: https://design-system.service.gov.uk/components/checkboxes/

### DIFF
--- a/libs/chlorophyll/src/lib/components/input/_checkbox.scss
+++ b/libs/chlorophyll/src/lib/components/input/_checkbox.scss
@@ -1,11 +1,14 @@
 @use '../../tokens';
 @use '../../common';
 @use 'sass:map';
+@use 'sass:math';
 
 /* Tokens todo: move/refactor */
 $background: tokens.$white;
 $background-hover: tokens.$white;
 $background-active: map-get(tokens.$blue, 1);
+$checkbox-size: 1.25rem;
+$sm-size-division-amount: 1.55;
 $checkmark-color: tokens.$white;
 
 /* Style for label */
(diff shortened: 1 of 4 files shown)


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we created for it [here](https://sebgroup.github.io/green/pr-271/chlorophyll/).

JohanObrink:
We always merge into `main` and then do releases from there.

I think some margins have to be adjusted in the large and possibly medium sizes. Otherwise looks good 👍🏻 

What do you think @hjalmers ?

image

hjalmers:
Yeah it looks like the alignment of the checkmark inside the checkbox needs some tweaking and also the checkmark transition/animation is a bit quirky: https://sebgroup.github.io/green/pr-271/chlorophyll/?path=/story/components-form-concepts-sizing--page

I've never seen any SEB app use different sizes of checkboxes and radio buttons, is it something that we want to support? From an accessibility perspective I guess it might be good to provide a bigger version but then again, everything should be bigger  and most likely the user should use zoom instead. If the interface is zoomed these large versions including the large radio button and checkbox will become gigantic compared to everything else on the page. Personally I think we should avoid using different sizes, so far I've only seen them being used for sign up forms etc in standalone settings.

mario-subo:
Thanks for the feedback! The largest one is indeed a bit wonky, will have another look at it. I'll also switch to `main` and adjust any discrepencies but before that maybe we can see if actually having bigger or smaller checkboxes and radio makes sense in the first place. @hjalmers raises some very good points. Honestly I like the idea of even regular sized checkboxes and radios being bigger (maybe the same height as a button).

Initially the idea for this PR came after playing around with the sizing part of the docs and I wanted to see what would happen if I add the size-lg to a checkbox input element.

mario-subo:
> @mario-subo are you still working on this or can we close it? I haven't seen this designed in Green so not sure that we need it?

Yeah we can close it. I don't think we need it.

PR #284, opened 2021-12-08
Feature/navbar

terrance456:


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we created for it [here](https://sebgroup.github.io/green/pr-284/chlorophyll/).

JohanObrink:
I like it 😄  For some reason the Storybook link always points to chlorophyll which is not built for this PR 😢  Maybe @hjalmers could have a look at that?

The correct link is: https://sebgroup.github.io/green/pr-284/react/?path=/story/components-navbar--navbar

My suggestion is to continue adding all the cases from the Chlorophyll storybook (transparent/dark/white) and setting the variant type in @sebgroup/extract so that it can be reused in Angular. After that we can merge. I think we will have to have a serious think about how to best add the children of the menu with automatic styling anyway.

Again: Great work 🎉 

PS: My OCD is telling me you should always have a blank line at the end of each file

terrance456:
Thanks for the feedback @JohanObrink! We will look into creating different variant for the navbar according Chlorophyll storybook as you mentioned and I believe there is still room for us to improve and we will continue to work on that as well.

libs/react/src/lib/navbar/navbar.tsx
```diff
@@ -0,0 +1,37 @@
...
+    const classNames: string[] = []
+    if (variant) classNames.push(variant)
+    setProps({ ...props, className: classNames.join(' ') })
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [variant])
+
+  return (
+    <nav role="navigation" {...props}>
+      <a
+        aria-label="Open seb.se in new tab"
```
kherP, line 23: it will be better if we allow users to set aria-label, href dynamically

libs/react/src/lib/navbar/navbar.tsx
```diff
@@ -0,0 +1,37 @@
...
+}
+
+//TODO: Add variants
+export const Navbar = ({ children, variant, title }: NavProps) => {
+  const [props, setProps] = useState<HTMLAttributes<HTMLElement>>({})
+  useEffect(() => {
+    const classNames: string[] = []
+    if (variant) classNames.push(variant)
+    setProps({ ...props, className: classNames.join(' ') })
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
kherP, line 17: probably we can use classnames or clsx library to join the class names

libs/react/src/lib/navbar/navbar.tsx
```diff
@@ -0,0 +1,37 @@
+import { NavbarVariant } from '@sebgroup/extract'
+import { HTMLAttributes, ReactNode, useEffect, useState } from 'react'
+
+interface NavProps {
```
kherP, line 4: should we extend nav props from default nav element?

JohanObrink:
Link component is merged into this branch making it possible to finalize navbar

PR #289, opened 2021-12-15
feat: 🎸 Added link component for green-react

ethan-qsh:


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we created for it [here](https://sebgroup.github.io/green/pr-289/chlorophyll/).

libs/react/src/lib/link/link.tsx
```diff
@@ -0,0 +1,21 @@
...
+  link?: string
+  text?: string
+}
+
+export const Link = ({ link, text }: LinkProps) => {
+  const props: HTMLProps<HTMLAnchorElement> = {}
+  if (link) props.href = link
+
+  return (
+    <div>
```
kherP, line 13: what is this div for?

libs/react/src/lib/link/link.tsx
```diff
@@ -0,0 +1,21 @@
+import React, { HTMLProps } from 'react'
+
+interface LinkProps {
+  link?: string
+  text?: string
+}
```
kherP, line 6: probably we should extend the props from default anchor element props, to support accessibility. For instance,
<Link link="some link" text="Read more" aria-label="Read more about this link" />.

libs/react/src/lib/link/link.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export const Link = ({ link, text }: LinkProps) => {
+  const props: HTMLProps<HTMLAnchorElement> = {}
+  if (link) props.href = link
+
+  return (
+    <div>
+      <a href={link} rel="noreferrer">
```
kherP, line 14: i dont think we should define `rel="noreferrer"` for the user

libs/react/src/lib/link/link.tsx
ethan-qsh: No purpose, we just overlooked it. We will remove it in the next commit.

libs/react/src/lib/link/link.tsx
ethan-qsh: Alright, will remove it in the next commit

libs/react/src/lib/link/link.tsx
ethan-qsh: Alright, do mean to add more of the default anchor element props to the interface?

kherP: my suggestion is to extend React.HTMLProps<HTMLAnchorElement> but maybe you can omit the children and href, depends on how you want this component to work

PR #290, opened 2021-12-15
Feature/link suggestion

JohanObrink:
I had a long meeting and couldn't resist doing some coding. I made quite big changes top the Link component trying to inject just the right amount of flexibility for the user and quality control for Green.

What do you think?


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we created for it [here](https://sebgroup.github.io/green/pr-290/chlorophyll/).

ethan-qsh:
It looks good! I like that it adds the option to change the link to a button and that it adds the role when it does that too. Should we be using the useState and useEffect hook to set the props for all the green-react components?

kherP:
> I had a long meeting and couldn't resist doing some coding. I made quite big changes top the Link component trying to inject just the right amount of flexibility for the user and quality control for Green.
> 
> What do you think?

Hi, haha 😄 sorry for being a little bit nosy, but setting classname in useEffect will cause some unnecessary flickering when user tries to use any style other than the default style. the reason is useEffect runs after each render. 

You can try in storybook by setting some default arguments in the url, i think you will see the link render in default style first then only the preset button style.

JohanObrink:
I just want to make sure the visuals update as properties update which they won’t when properties are calculated (as far as I understand). One way to solve it is to break out the calculation function from the useEffect and use it both in useEffect and in the initial useState.

/Johan

kherP:
Or we can use some [helper ](https://github.com/lukeed/clsx)library to construct the classname
```
<a {...props} className={clsx({ [button]: typeof button === 'string', 'button': button })}>{ children }</a>
```
just some suggestions

JohanObrink:
Two things:

1. You do not want to run a lambda function in the render as it will make it pre-render all the time
2. If we construct a function that builds element properties from component properties, we can put it in extract and reuse it across frameworks (even if we need to convert between class/className)

Also: @hjalmers - should we remove the class=”button” and replace it with role=”button”?

/Johan

kherP:
i see, do you mind to elaborate on the impact of running a function during render?

JohanObrink:
From what I understand, a lambda function in the render method prevents React from detecting if things changed which leads to the component being pre-rendered all the time.

hjalmers:
@JohanObrink regarding class vs role - role="button" should be used to let the browser and assertive technologies know that the element is a button and should behave as one. In this case I guess it doesn't, unless we tie it to something else than an anchor element (which is used by and for links), and at the same time add javascript to make it behave like a button. E.g `<div role="button">I'm not a button but should behave like one so please add listeners and styling</div>`.

If we want a link to behave like a button (not just look like one) we should use `<button>` or `<input type="button">` instead. Classes are just for appearances so I think it makes sense to use class="button" just to make it look like a button, assertive technologies and screens readers will still se it for what it is (a link).

As a side note, personally I'm not a big fan of wrapping native html elements in web components (native, angular, vue, react, svelte etc) just for the sake of it unless they add some logic or smartness. It's a matter of preference of course but if all the inputs/props are just used to set attributes on native elements it tends to add an unnecessary layer of complexity. 

E.g:

**Native**
<a href="http://seb.se" target="_blank" rel="external">Go to SEB.se <i class="some-special-icon"/></a>

**Component**
<Link href="http://seb.se" target="_blank" rel="external">Go to SEB.se <i class="some-special-icon"/></Link>

Apart from not adding an additional layer which has to be maintained, the native approach will render without JavaScript which is an added bonus:)

Most of the styling will probably be determined by where the link is contextually placed in the DOM, eg. inside an alert or card, or by the target and optionally class or rel (relationship). I.e. in the example above the icon could be omitted and added either using a class or preferably using rel="external":)

JohanObrink:
There are cases where a link works like a button. F.e when an OK button should just take you to the next step in a process expressed as a URL. The question is wether we should use thinks that look like buttons but aren’t described as such.

When it comes to basic components rendered through React, I would say it is not only possible but even recommended. In React, everything is always rendered by JS anyway and wrapping atom components gives control over exposed properties and their possible values as well as hiding the implementation details from the user of the component. This enforces code standards without having to teach developers about them.

hjalmers:
Well if the next step in a process is just a link with a reference to another page or fragment, keeping the role="link" (default behaviour for anchor elements) and treating it as a link is still valid. If the anchor element on the other hand should trigger something else like running a custom function to save a state then role="button" could be more appropriate. Anchor tags and the role="link" will just tell browsers and assistive technologies pressing enter will follow the hyperlink (e.g taking the user to the next step). Using role="button" for anchor elements make sense for scenarios when you use <a href="#"> and prevent the default behaviour which is to follow the hyperlink, then again it's probably better and more correct to use a button when you omit the href or prevent the default behaviour for a link. 

In any case I'd put my vote on using class for appearance here as we'll probably have the scenario of making a normal link look like a button without changing the meaning or expected behaviour. Using a class wouldn't have to restrict the use of role="button" either but then we should probably use a button too in the first place? Same goes for the other way around of styling a button like a link using the role="link":)

Yeah it's a matter of taste I guess, and if you want to add type checking or restrict possible options then wrapping native elements would be the way to go. Most IDEs provide code completion and hints for native html elements too and warn you if you use attributes incorrectly:)

ethan-qsh:
Based on your comments @JohanObrink and @hjalmers, should this component still be built or do we want to omit it in favor of the user using a native html anchor element?

JohanObrink:
@ethan-qsh Oh it should be built... at least in React. We'll have to see about Angular.

It's important to realise that React is a Component rendering system while Angular is a web framework. It means 100% feature parity would unavoidably violate the principles of at least one and probably both of them.

PR #304, opened 2021-12-23
Feature/list

ethan-qsh:


libs/react/src/lib/list/listItem.tsx
```diff
@@ -0,0 +1,17 @@
...
+  listType?: ListType
+  children?: ReactNode
+}
+
+const ListItem = ({ listType, children, ...props }: ListItemProps) => {
+  if (listType === 'table') {
+    return <div>{children}</div>
+  } else {
+    return <li {...props}>{children}</li>
+  }
```
mario-subo, line 14: I guess you will have to add another condition to wrap the children in `<dl>` as well

mario-subo:
Note: I've added re #314 and re #315 which relate to the list but the style will probably have to be fixed in chlorophyll

libs/react/src/lib/list/listItem.tsx
ethan-qsh: Yeap, will have to look into how to implement that. The `<div>` on line 11 is used to style the table that's why its there

ethan-qsh: I made the changes and I saw that the tag was changed from `<dl>` to `<dd>` in another PR so I made the changes accordingly

hjalmers (approved):

PR #305, opened 2022-01-03
fix: 🐛 fixed default modal storybook

terrance456:


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we created for it [here](https://sebgroup.github.io/green/pr-305/chlorophyll/).

ethan-qsh:
Terrance and I noticed that the SlideIn modal and the Takeover modal don't have any styles yet so they look bad in the storybook. Are we planning to follow a specific design guideline for this component? If yes, can you link us in that direction? Thanks!

@JohanObrink @hjalmers

PR #328, opened 2022-01-12
Feature/dropdown angular

Maximilian-R:


JohanObrink:
@Maximilian-R Nice to see that you started refactoring the Angular version of the Dropdown. I think the best way to measure progress on the issue would be to copy and convert the component tests from the React version. That way we know for a fact which parts do and don't work. Since we need the tests before merge anyway, might as well use them as a support while developing 😄

PR #332, opened 2022-01-13
fix: 🐛 Fixed merge conflict in list/_mixins.scss

JohanObrink:


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we created for it [here](https://sebgroup.github.io/green/pr-332/chlorophyll/).

libs/chlorophyll/scss/components/list/_mixins.scss
```diff
@@ -73,7 +65,7 @@ $border-color: tokens.$border-color;
       flex: 1 1 auto;
     }
 
-    dd {
+    dl {
```
hjalmers, line 68: It should be like this here
```suggestion
    dd {
```

hjalmers (changes requested):

PR #343, opened 2022-01-14
refactor: 💡 dropdown storybook

ethan-qsh:


hjalmers (commented):
Might be something from extract and not specific for the react implementation but we should support Enter key as well for expanding/collapsing dropdown. https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html 

Escape key should also collapse the dropdown without updating the selected value.

PR #408, opened 2022-02-16
Refactor/405 react input component

ethan-qsh:
Created validators for the component

image


hjalmers (approved):
LGTM, have you thought about how validation rules should be declared? Not sure about the typical way of doing it with react but would be nice to have or provide something similar to what's offered by reactive forms and the form builder in angular. Eg.

```ts
this._fb.group({
            name: ['', [Validators.required]],
            age: [18, [Validators.required, Validators.min(18), Validators.foo]],
            email: ['', [Validators.required, Validators.email]]
        })
`
```

The above will in angular generate a form with a couple of requirements that in turn will be emitted in an error object if not fulfilled. That error object can be used to populate error texts with the option to pass custom texts using map, eg. 

```json
{
  min: 'Value has to be greater than or equal to 18',
  required: 'Field is required',
  email: 'Incorrect email address',
  foo: 'Foo validator returned some custom error'
}
```

PR #413, opened 2022-02-17
build: fix dependencies in package.json for libs

### LINKED ISSUE #412: Incorrect dependencies for publishable libs
NX provides a feature that analyses the code and determines which dependencies to add to the package.json when building libraries. Although it's a nice feature and good in theory, it's currently a bit strict and  in some cases we want it to allow for floating ranges etc.

One example being determining angular versions to use. If undefined in package.json of lib, nx will correctly add it but does so with a fixed version e.g. 

```
"@angular/forms": "^13.2.3"
```

This is good and promotes keeping dependencies up-to-date but it's also a bit strict in some cases and sometimes we might want something like:

```
"@angular/forms": "^12.0.0 || ^13.0.0"
```

Some dependencies are also unwanted/need like react for an angular lib when we're only relying on typings from @types/react, or rxjs for react if not used. 

To have more control of which dependencies should be added (and what version/range),`updateBuildableProjectDepsInPackageJson` option in project config for build can be set to false.

hjalmers:
closes #412

### DIFF
--- a/libs/angular-charts/package.json
+++ b/libs/angular-charts/package.json
@@ -3,7 +3,10 @@
     "version": "1.0.0-beta.5",
     "peerDependencies": {
         "@angular/common": "^12.0.0 || ^13.0.0",
-        "@angular/core": "^12.0.0 || ^13.0.0"
+        "@angular/core": "^12.0.0 || ^13.0.0",
+        "@angular/platform-browser": "^12.0.0 || ^13.0.0",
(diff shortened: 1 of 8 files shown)


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we created for it [here](https://sebgroup.github.io/green/pr-413/chlorophyll/).

PR #419, opened 2022-02-25
Feature/417 form validator react

ethan-qsh:
Added the barebone validator for React Form component (right now it is just the email and it gives the invalid warning when you input anything into the email input) but I wanted to start a discussion here on:

1. The regex structure for the inputs that we want to validate 
2. If handling the changes to the inputs in both the form component and input component should be the way to go or should it be only in one component (I think it should be in one component, the form component)


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we created for it [here](https://sebgroup.github.io/green/pr-419/chlorophyll/).

terrance456:
> Added the barebone validator for React Form component (right now it is just the email and it gives the invalid warning when you input anything into the email input) but I wanted to start a discussion here on:
> 
> 1. The regex structure for the inputs that we want to validate
> 2. If handling the changes to the inputs in both the form component and input component should be the way to go or should it be only in one component (I think it should be in one component, the form component)

What I think is, the component should only has the indicator (which conditional render whenever there is error or success), where this indicator should be handle by the developer when they are using the component, cause there can be multiple condition for forms, I don't think we can handle all the validation for the form conditions, what we can do is provide a utility function which checks the input values based on that the dev can toggle the indicator. This utility function can validate the input values base on few condition eg: emails, required, maxLength, minLength. What you guys think? @ethan-qsh  @JohanObrink  @hjalmers  @mario-subo

terrance456:
Built most simplest form validator and only works with text inputs for now, what you guys think ? @JohanObrink @hjalmers @mario-subo @ethan-qsh

libs/react/src/lib/form/formContext.tsx
```diff
@@ -0,0 +1,76 @@
...
+      const resetErrors = () => {
+        setErrors({})
+      }
+      currentRef.addEventListener('reset', resetErrors)
+
+      return () => currentRef.removeEventListener('reset', resetErrors)
+    } else {
+      // eslint-disable-next-line @typescript-eslint/no-empty-function
+      return () => {}
+    }
```
eweseong, line 59: the `else` statement does not seemed to be doing anything, perhaps we can remove it?

libs/react/src/lib/form/formContext.tsx
```diff
@@ -0,0 +1,76 @@
...
+  React.useEffect(() => {
+    if (formRef.current) {
+      const currentRef: HTMLFormElement = formRef.current
+      const resetErrors = () => {
+        setErrors({})
+      }
+      currentRef.addEventListener('reset', resetErrors)
+
+      return () => currentRef.removeEventListener('reset', resetErrors)
+    } else {
```
eweseong, line 56: maybe we can refactor this into `onReset` event handler?

libs/react/src/lib/form/formContext.tsx
terrance456: Agreed, but the console seems to throw error saying not all path returns maybe something to do with the `@typescript-eslint`

libs/react/src/lib/form/formContext.tsx
terrance456: Yeap will refactor that

libs/extract/src/lib/helperFunction.ts
```diff
@@ -10,3 +12,58 @@ export const validateClassName = (value: IndicatorType) => {
...
+          setErrorRemove(setError, name)
+          return
+        }
+      }
+      default: {
+        return
+      }
+    }
+  }
+}
```
eweseong, line 47: looking at the parameter it seemed that the function has many responsibility, maybe we can decouple the error dispatch to the caller and let the function focuses more on validation?

```typescript
export type Rule = {
  custom?: (...args: Array<any>) => string; // expect return value to be error code or empty string
  type?: string;
  // future condition such as min, max, length etc.
}

export const validateInputValue = (name: string, rule: Rule, value: string) => {
  if (rule?.custom) {
    return rules.custom();
  }

  switch (rule?.type) {
    case 'Required': {
      return (value === '' || value === undefined || value === null) ? 'error_code_empty';
    default: 
      return '';
   }
}
```

terrance456: Yeap, this looks way cleaner and readable, will refactor it

libs/extract/src/lib/helperFunction.ts
```diff
@@ -1,3 +1,5 @@
+import * as React from 'react'
+
 export type IndicatorType = 'valid' | 'invalid' | 'neutral'
```
kherP, line 3: maybe we can replace neutral with disabled as if the form is neutral, means it is valid

libs/react/src/lib/form/input/input.tsx
```diff
@@ -9,25 +9,28 @@ import {
   TextInputProps,
 } from '../types'
 import { IndicatorType, validateClassName } from '@sebgroup/extract'
+import React from 'react'
```
kherP, line 13: maybe we can remove this component as we are using formItem to wrap the input. so when user use this form element, they will use with the other components with green/react. i.e:

```
<Form>
  <FormItem>
    <Email />
  </FormItem>
   ...
</Form>
```

libs/react/src/lib/form/formContext.tsx
```diff
@@ -0,0 +1,76 @@
...
+    if (fields) {
+      const newErrors: Array<string | undefined> = []
+      Object.keys(fields).map((keys: string) =>
+        newErrors.push(
+          validateInputValue(keys, fields[keys], values?.[keys], setErrors)
+        )
+      )
+      if (!newErrors.includes('error')) {
+        console.log('submitted')
+      }
```
eweseong, line 43: combine with the suggestion above, we can refactor this section into
```typescript
const hasError: boolean = false;

Object.keys(fields).forEach((key: string) => {
  const errorMessage = validateInputValue(key, fields[key], values?.[key]);
  hasError = hasError || !!errorMessage;
  setErrors((prevErrors: Record<string, string>) => ({
    ...prevErrors,
    [key]: errorMessage ?? undefined
  }));
});

if (!hasError) {
  console.log('submitted')
}
```

libs/react/src/lib/form/formContext.tsx
eweseong: is the eslint configure to support react and react hooks?

libs/extract/src/lib/helperFunction.ts
terrance456: The neutral actual displays error message without any specific color or border, so I renamed to it something better ?

libs/react/src/lib/form/input/input.tsx
terrance456: Remove which component ? Input component ? I only see this 
Capture

libs/react/src/lib/form/formContext.tsx
terrance456: Cool, will refactor it accordingly

libs/react/src/lib/form/formContext.tsx
terrance456: > is the eslint configure to support react and react hooks?

Not too sure about this, anyways I removed the event listener and pass the callback to form reset event handler.

libs/extract/src/lib/helperFunction.ts
eweseong: `info`?

terrance456: Yeap sounds good to me

libs/react/src/lib/form/formContext.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+    if (!hasError) {
+      onFormSubmit && onFormSubmit(values)
+    }
+  }
+
+  const onResetForm = () => {
+    setValues({})
+    setErrors({})
+  }
```
eweseong, line 41: let's be consistent with event handler naming, if the `on` prefix is decided for the project, add to all; otherwise if it isn't maybe we can use the following naming convention
> If it is a function or method, use verbs like getName, getTotalCount, setProductName and addCustomerName etc.

ie: `submitForm`, `resetForm`

libs/react/src/lib/form/formItems.tsx
```diff
@@ -0,0 +1,48 @@
...
+  name,
+}) => {
+  const { setValues, setErrors, setFields, errors } = useFormContext()
+
+  React.useEffect(() => {
+    setFields((fields: Record<string, any>) => ({
+      ...fields,
+      [name]: validate?.rules,
+    }))
+  }, [])
```
eweseong, line 25: given that we are making side effect changes onto context, was wondering if it make sense to undo the side effect when the component is being unmounted? similar to how we remove event listener when component is unmounted.

libs/react/src/lib/form/formContext.tsx
terrance456: Sure will refactor it

libs/react/src/lib/form/formItems.tsx
terrance456: Not too sure about this, should I be returning something here i.e : `return () =>  ....` ?

eweseong: yes, let say if the `FormItems` was dynamically removed, should the `fields` from form context still retain the rules and errors for it?

terrance456: Ahh correct, I should remove it whenever the component unmount

hjalmers:
Hmm, yeah since I haven't worked anything with forms in react I'm not sure what most teams and developers prefer and use, but I think it makes sense not to add too much logic to each form "item". Then again, if we have a component for email it would make sense to add basic email validations to it. 

I like the way reactive forms handles form validations in angular i.e something like:

```ts
this._fb.group({
            name: ['', [Validators.required]],
            age: [18, [Validators.required, Validators.min(18), Validators.foo]],
            email: ['', [Validators.required, Validators.email]]
        })
```

You specify the rules in the component holding the form (not in each form element although it would be possible and might make sense for some like email, personal identity number etc.). The above will in angular generate a form with a couple of requirements that in turn will be emitted in an error object if not fulfilled. That error object can be used to populate error texts with the option to pass custom texts using a map, eg. 

```js
{
  min: 'Value has to be greater than or equal to 18',
  required: 'Field is required',
  email: 'Incorrect email address',
  foo: 'Foo validator returned some custom error'
}
```

Common validators for `required`, `email`, `min`, `max`, `minLength`, `maxLength` etc. are provided by default and can easily be extended and it's easy to create custom validators as well. Would be nice if we could try and use the same terms and behaviour for both react and angular and define custom validator functions in extract. Also only on error should be displayed for a field at any given time which allows us to keep the props for each form item pretty slim i.e. only provide one for message and if it's valid or not etc.

terrance456:
> Hmm, yeah since I haven't worked anything with forms in react I'm not sure what most teams and developers prefer and use, but I think it makes sense not to add too much logic to each form "item". Then again, if we have a component for email it would make sense to add basic email validations to it.
> 
> I like the way reactive forms handles form validations in angular i.e something like:
> 
> ```ts
> this._fb.group({
>             name: ['', [Validators.required]],
>             age: [18, [Validators.required, Validators.min(18), Validators.foo]],
>             email: ['', [Validators.required, Validators.email]]
>         })
> ```
> 
> You specify the rules in the component holding the form (not in each form element although it would be possible and might make sense for some like email, personal identity number etc.). The above will in angular generate a form with a couple of requirements that in turn will be emitted in an error object if not fulfilled. That error object can be used to populate error texts with the option to pass custom texts using a map, eg.
> 
> ```js
> {
>   min: 'Value has to be greater than or equal to 18',
>   required: 'Field is required',
>   email: 'Incorrect email address',
>   foo: 'Foo validator returned some custom error'
> }
> ```
> 
> Common validators for `required`, `email`, `min`, `max`, `minLength`, `maxLength` etc. are provided by default and can easily be extended and it's easy to create custom validators as well. Would be nice if we could try and use the same terms and behaviour for both react and angular and define custom validator functions in extract. Also only on error should be displayed for a field at any given time which allows us to keep the props for each form item pretty slim i.e. only provide one for message and if it's valid or not etc.

Yeap agreed, we should have both react and angular and define custom validator functions in extract so that we can reuse in both but I think we can have `required` rule for now, if things are good then we can improvise it, I took the inspiration from here https://ant.design/components/form where we pass in rules for every input component with custom error message, where the developer will be flexible to pass in a custom validation callback to validate the input.

PR #428, opened 2022-03-03
chore(deps-dev): bump

hjalmers:


JohanObrink:
@hjalmers Stylelint was complaining about missing `postcss-scss` plugin so I added it to dev-deps and it worked. Can you validate on your end before merging?

Issue #449, opened 2022-03-31
Prevent background scroll for dropdown


hjalmers:
Re-add body-scroll lock to prevent background scrolling for dropdown on mobiles.

Maximilian-R:
Found some more issues while looking at this, will include a fix for them too:

- Limit listbox height so it cannot be larger than the screen
- Re-add Chevron/Arrow blue color
- Limit toggle text to one line
- Add close button background color css variable (only defined in alert component)

Maximilian-R:
https://github.com/sebgroup/green/blob/34a2664128fe2d60e27aacdd4bbdf52fbdba2ed0/libs/chlorophyll/scss/components/close/_mixins.scss#L25

The close button in dropdown slide-up modal uses --sg-hsl-contrast which is not included in the light theme from @use '@sebgroup/chlorophyll/scss/components/use-green-scope'; 

I'm a bit lost in the color/theme setup, could you have a quick look? :) 

@hjalmers

hjalmers:
@Maximilian-R yeah it's a bit messy at the moment:( I missed this comment yesterday but will take a look at it when I get a chance:)

PR #453, opened 2022-04-05
fix(dropdown): body-scroll-lock and styling

Maximilian-R:
* Body scroll lock when dropdown is open
* Toggle-button text overflow
* Listbox & modal max height
* Chevron blue color
* Use initial values to get correct styling for listbox before dropdown is created
* Toggle-button cursor
* Close button

### DIFF
--- a/apps/angular-lib-dev/src/app/dropdown/dropdown.component.html
+++ b/apps/angular-lib-dev/src/app/dropdown/dropdown.component.html
@@ -53,3 +53,34 @@
 </div>
 <div>Value: {{ dropdownValue3 | json }}</div>
 <div>Value: {{ dropdownValue4 | json }}</div>
+
+<h3 class="my-4">Form group and validation</h3>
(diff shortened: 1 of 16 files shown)


Maximilian-R:
related to #449

sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybooks we've created for it: <ul><li><a target="_blank" href="https://sebgroup.github.io/green/pr-453/chlorophyll/">chlorophyll</a></li><li><a target="_blank" href="https://sebgroup.github.io/green/pr-453/angular/">angular</a></li></ul>

hjalmers (approved):
LGTM! Perhaps we should just disable scroll on small screens (when scrolling in the background is an issue)? That could be fixed in another PR though:)

I fixed a small issue with the updated css which caused the chlorophyll example to break.

I also added support for adding a label and experimental support for validation. I think we should create a form-info component for the "hint" and possible errors so this implementation is just temporary.

PR #492, opened 2022-05-03
Sjkayle feature/badge component

JohanObrink:
Moved title in React component to children


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybooks we've created for it: <ul><li><a target="_blank" href="https://sebgroup.github.io/green/pr-492/react/">react</a></li><li><a target="_blank" href="https://sebgroup.github.io/green/pr-492/angular/">angular</a></li></ul>

hjalmers:
@JohanObrink and @sjkayle I took the liberty to refactor the angular version a bit. Mainly shifting from using a typical component selector `<ngg-badge>` to an attribute selector `ngg-badge` so instead of this:

```html
<ngg-badge>Text</ngg-badge>
```
You now use it like this:

```html
<span ngg-badge>Text</span>
```

The main reason for this is that Angular wraps the html in a custom selector (just like other web components). One benefit with this is that it allows angular to scope styles to individual components with and without the use of shadow dom, it also allows angular components to be compiled to native web components. One drawback is that it affects the use of standard css rules on these components (in some cases).

Consider the following css rule:

```css
.badge + .badge {
  margin-left: 1rem;
}

```

The following markup would produce two badges and the second badge would have left-margin set to 1 rem since our css rule states that .badge preceded by .badge should have a left margin.

```html
<span class="badge">Text</span>
<span class="badge">Text</span>
```

Using a standard component selector with angular these rules wouldn't be applied as the result would look like this:

```html
<ngg-badge><span class="badge">Text</span></ngg-badge>
<ngg-badge><span class="badge">Text</span></ngg-badge>
```

One final note is that I think we have to consider how badges are used. For example how will the consumer know that a badge has been removed now as close will just remove it from the dom? Perhaps it should be implemented as a form control, i.e.expose an array of badges were the state could be tracked by looking at the state of the form control (similar to a multi select). 

If you're not borred already and want to keep on reading;)

### Some thoughts on creating components

Styling components with angular is typically not a problem as these kind of css rules are often applied to native elements i.e. buttons, anchors (links), divs etc since we don't write css for web components that are meant to be self contained (they should include their own logic, styling and expose css variables to allow custom theming). Personally I tend to avoid wrapping native components for this reason. E.g. creating a button component for a button doesn't really make sense if we're only interested in controlling which attributes (like classes) that should applied to the element (that's what directives are for). With directives we can extend the behaviour for native elements. If we on the other hand want to manipulate the markup of native elements such as adding other components or child elements or add additional logic, we have the option to use other selectors to create/extend existing ones. eg. instead of creating a new button component using `<ngg-button>` we could use a class or an attribute as selector to alterer or extend the native behaviour as well as its' markup. You typically create angular components for things that natively doesn't exist or when you invent something new (eg. dropdown, datepicker etc.).

Issue #490, opened 2022-04-29
Card Component


sjkayle:
Card component for Angular

Todos:
- [ ] Markup
- [x] Story

PR #497, opened 2022-05-06
feat: 🎸 new angular card component

sjkayle:
✅ Closes: #490


sjkayle:
Hi @hjalmers, do you mind explaining the difference of the components in `~/libs/angular/src` and `~/apps/angular-lib-dev`? :)

sjkayle:
@hjalmers Hi Robert, will have to refactor the way I created this card component based on your comment from my previous pull request. Thank you! :)

hjalmers:
@sjkayle are you using this component anywhere currently? I think makes sense to add the things you need to improve developer experience or to solve a specific problem related to using a component. Personally I think the card component is a bit tricky as it's hard to explain it as a component without a specific use case.

I.e. if it's just a wrapper for the layout e.g.,

```html
<card-component></card>
```

And you pass in content using slots like this:

```html
<card-component>
  <header>Header</header>
  <main>Body content</main>
  <footer>Footer</footer>
</card-component>
```

It's basically just the same as doing this:

```html
<section class="card">
  <header>Header</header>
  <main>Body content</main>
  <footer>Footer</footer>
</section>
```

The only real benefit of wrapping it is that we can change the some aspects of the card without the consumer having to update their markup. On the other hand we also add an additional layer which adds complexity. Structural changes to the component will most like affect the component either way, either by introducing new attributes (props and inputs) or how slots are used.

Another benefit could be if we add logic to it which otherwise would have to be implemented outside it. Eg. ability to collapse  or something else which requires javascript.

Since cards typically are used for layout, it might make sense to either use a directive (for controlling aspects of the card element) or an attribute selector if we add more logic to it. Or perhaps a "cards" component which organises cards in a grid. But then again it depends a bit on the problem you trying to solve or the developer experience you want to enhance:)

sjkayle:
@hjalmers Understood. Thank you very much for your feedback. :)

PR #506, opened 2022-05-19
Hotfix/dropdown select value

hjalmers:


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we've created for it: <ul><li><a target="_blank" href="https://sebgroup.github.io/green/pr-506/angular/">angular</a></li></ul>

libs/angular/src/lib/dropdown/dropdown.component.ts
```diff
@@ -299,13 +298,16 @@ export class NggDropdownComponent
...
 
   private setSelectionByValue(value: any) {
-    if (this._value !== value) {
-      this._value = value
-
-      const selected = this.handler?.dropdown?.options.find(
-        (option) => option.value === value
-      )
-      if (selected) this.handler?.select(selected)
+    if (!this.dropdown?.isMultiSelect) {
```
Maximilian-R, line 301: Should the multiselect also be able to set selection by value input?

hjalmers: @Maximilian-R that was my intention to begin with but it was a bit tricky to implement with the current setup so decided to skip that use case for now as it's possible to work around it by using the selected property:)

PR #507, opened 2022-05-24
Feature/ng alert

sjkayle:


hjalmers:
@sjkayle do you have an example where you use this as well? It's always nice with actual use cases and examples (something I think we should try and add to storybook too).

I'm thinking a bit about the state as well and how it's handled. Since it's removed internally now from within the alert we should emit an event once it's removed to notify the view and other components who might want to know the state or what do you think?

For components that are fixed and "dumb" I typically tend to leave the state out eg.

```html
<form>
  ...
  <error *ngIf="hasError$ | async"></error>
</form>
```

The form or the component holding the form can control the state of the error using `hasError` which could be synced using some state management framework like ngrx, akita or a regular service with getters and setters or rxjs and subjects. If we don't want to complicate things, using an output for close could be one way to go to handle close action without introducing a service that would have to be provided somewhere. 

Another typical scenario is that we want to display an alert instead of something else which could be achieved like this (one of many ways).

```html
<foo *ngIf="hasBar; else error"></foo>
<ng-template #error>
    <error></error>
</ng-template>
```

Another approach  could be to pass the state using an input and use *ngIf inside the component, that would however result in a "empty" component that's not going to be destroyed or removed correctly from the DOM when state changes which we typically want.

In any case we need a way to sync the state with the view or component using the alert. It could be done or achieved in multiple ways and I think we need to have at least one real use case for when we'd want to show an alert and what things we need to support. If you dismiss the alert now by clicking close and then reload the page it would show up again and currently there's now way to handle the state, at least not in a way that's straightforward.

PR #543, opened 2022-07-15
New component in Chlorophyll: Slide toggle

splashdust:


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we've created for it: <ul><li><a target="_blank" href="https://sebgroup.github.io/green/pr-543/chlorophyll/">chlorophyll</a></li></ul>

hjalmers:
Looks good 👍  have you considered the role [switch](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/switch_role)? Perhaps we should have an example where it's placed in a form group too? Mainly for the purpose of laying it out nicely if used in forms, possible even handle additional information or validation text. I.e. something like this:

```html
<div class="form-group">
    <span id="cookies-label-1">Enable cookies</span>
    <span class="form-info" id="formInfo1">Some info or help</span>
    <button class="slide-toggle" aria-pressed="false" onClick={toggle} aria-labelledby="cookies-label-1"></button>
    <span class="form-info">Hint or error</span>
</div>
```

splashdust:
Thanks! I'll add the form group example and have a closer look at the `switch` role to see if that makes more sense than `aria-pressed`.

vsjolander (approved):
Awesome! Thanks for contributing 🤙

PR #545, opened 2022-07-18
fix: 🐛 Configurable font path when setting up use-green-scope

vsjolander:
When setting up use-green scope it was not possible to configure font path. Now you can do it by adding a config to the import of use-green-scope


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we've created for it: <ul><li><a target="_blank" href="https://sebgroup.github.io/green/pr-545/chlorophyll/">chlorophyll</a></li></ul>

PR #547, opened 2022-07-21
Feature/ngg toggle selector

sjkayle:
Closes [#551](https://github.com/sebgroup/green/issues/551)

A new toggle selector component for green angular with additional feature to add description to each option item.

This PR also resolves bug fixes from the toggle selector component existing in `ng-components`:
- Value not being returned when multi-select is enabled
- Throwing warning on option item click when multi-select is enabled


libs/angular/src/lib/toggle-selector/toggle-selector.component.html
```diff
@@ -0,0 +1,48 @@
...
+      [disabled]="disabled || item.optionItem.disabled"
+      (focus)="focus(idx)"
+      (blur)="blur()"
+      [defaultChecked]="item.selected"
+    />
+    <div
+      class="icon-container"
+      [class.icon-only]="!item.optionItem.label"
+      *ngIf="item.optionItem.icon"
+      [innerHTML]="item.optionItem.icon | safeHtml"
```
splashdust, line 29: Is there some way we can avoid using unsanitized html here? Especially considering this value comes from an input that could potentially contain compromised data.
I think we should either limit icon usage to a predefined list, or write a custom sanitizer that only allows whatever html is required to render icons.

sjkayle:
Hi @splashdust, as per your comment, kindly review my latest commit if this could be an acceptable solution? :D Thank you.

sjkayle:
> 

Hi @splashdust, do you mean I should import this [DOMPurify](https://github.com/cure53/DOMPurify) library to our green project?

splashdust:
@sjkayle Yes, I think it makes sense to add DOMPurify as a dependency for this use case. At least until the [Sanitizer API web standard](https://wicg.github.io/sanitizer-api/) becomes commonly available in all major browsers.

DOMPurify is only 25kb or so minimized, so I think it's worth it for the added security. And we'll probably need it in other components too eventually.

sjkayle:
@splashdust Hello, I am having trouble adding the DOMPurify package to our green project.

image

Could you kindly advise? :D Can I try using `npm` instead of yarn?

vsjolander:
@sjkayle is this being worked on or should we close this PR?

sjkayle:
@vsjolander I believe this PR could be closed at this time. I have read from SEB's Design Library that the Toggle Selector component would be deprecated in the future anyway.

PR #553, opened 2022-08-04
Fix/dropdown scroll

### LINKED ISSUE #219: Allow custom template for dropdown using slot
It should be possible to us a custom template for dropdowns something like:

```html
<ngg-dropdown [options]="options" [control]="control">
  <div>
    <span class="label">{{option.label}}</span>
    <span class="value">{{option.value}}</span>
  </div>
</ngg-dropdown>
```

### LINKED ISSUE #463: Scroll position for dropdown list not updated with keyboard navigation
**Issue:**
When navigating with arrow keys in long lists with scroll, focus is not moved (scroll position not updated), see screenshot below.

**Desired behavior:**
Using arrow keys should set focus to scroll the container when focus shifts.

image

Maximilian-R:
closes #463
closes #219

### DIFF
--- a/apps/angular-lib-dev/src/app/dropdown/dropdown.component.html
+++ b/apps/angular-lib-dev/src/app/dropdown/dropdown.component.html
@@ -53,7 +53,14 @@
         display="name"
         [texts]="texts"
         [(value)]="dropdownMultiValue2"
-      ></ngg-dropdown>
+      >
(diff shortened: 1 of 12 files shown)


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we've created for it: <ul><li><a target="_blank" href="https://sebgroup.github.io/green/pr-553/chlorophyll/">chlorophyll</a></li></ul>

PR #590, opened 2022-09-23
Feat/chart tooltip number format

### LINKED ISSUE #575: Tooltip value (numbers) need to have white space for large numbers
The problem is that the tool-tip value doesn't show very nice for larger numbers. We need green help to fix this. As pic below shows: 
**The value of tool-tip need to be in a prettier format like this with white spaces: 312 520. For larger number needs to be like 23 345 123 kr.**

### A suggestion of how solution looks like:
The template of tool-tip is like this today: 
```
const tmplTooltip = `
  <dl class="bb-tooltip-list">
    <dt>{=TITLE}</dt>
    {{
    <dd class="{=CLASS_TOOLTIP_NAME}" style="--color: {=COLOR}">
      <span>{=NAME} {=VALUE}</span>
    </dd>
    }}
  </dl>
`;
```

A func to fix this problem can be like this: 
```
numberWithSpaces(x) {
    return x.toString().replace(/\B(?=(\d{3})+(?!\d))/g, " ");
  }
```

Maybe Green can modify template and add this function instead? 

```
const tmplTooltip = `
  <dl class="bb-tooltip-list">
    <dt>{=TITLE}</dt>
    {{
    <dd class="{=CLASS_TOOLTIP_NAME}" style="--color: {=COLOR}">
      <span>{=NAME} {=numberWithSpaces(VALUE)}</span>
    </dd>
    }}
  </dl>

```

splashdust:
closes #575

### DIFF
--- a/libs/angular-charts/src/lib/chart/chart.stories.ts
+++ b/libs/angular-charts/src/lib/chart/chart.stories.ts
@@ -57,7 +57,7 @@ SimpleBar.args = {
           label: 'Antal',
         },
       },
-      tooltipNumberFormat: value => `${value} kr`
+      tooltipNumberFormat: value => `${value} st`
     },
   },
 }
--- a/libs/angular-charts/src/lib/chart/chart.stories.ts
+++ b/libs/angular-charts/src/lib/chart/chart.stories.ts
@@ -57,6 +57,7 @@ SimpleBar.args = {
           label: 'Antal',
         },
       },
+      tooltipNumberFormat: value => `${value} kr`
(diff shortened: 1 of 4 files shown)


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we've created for it: <ul><li><a target="_blank" href="https://sebgroup.github.io/green/pr-590/angular-charts/">angular-charts</a></li></ul>

Maximilian-R (approved):

PR #594, opened 2022-10-03
feat: 🎸 Modal (Angular)

SEBNiklas:
Modal component (Angular).

Issues
- In IBP the backdrop is not visible (because the .backdrop class collides with a class with the same name in IBP). 
- Placement of footer buttons for mobile (small devices) is not the same as the sketches in design library


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we've created for it: <ul><li><a target="_blank" href="https://sebgroup.github.io/green/pr-594/angular/">angular</a></li></ul>

libs/angular/src/lib/modal/modal.service.ts
```diff
@@ -0,0 +1,44 @@
...
+    constructor (@Inject(DOCUMENT) private document: Document) {}
+
+    public lockBody(htmlElement: HTMLElement) {
+        // Check if lock already exists
+        if (this.locks.some((l) => l === htmlElement))
+            return;
+
+        // If no locks exists, then add class
+        if (this.locks.length <= 0) {
+            this.document.body.classList.add("ngg-overflow-hidden");
```
Maximilian-R, line 19: I cannot find any implementation of this class? Anyhows, since locking the body in different browsers/devices is a bit different I think we should go with a library to handle this. For the dropdown and datepicker we have used 'body-scroll-lock'. It should be an easy replacement :)

libs/angular/src/lib/modal/modal.module.ts
```diff
@@ -0,0 +1,18 @@
+
+import { CommonModule } from '@angular/common';
+import { NgModule } from '@angular/core';
+import { NggModalBodyComponent, NggModalComponent, NggModalFooterComponent, NggModalHeaderComponent } from './modal.component';
+
+const EXPORTED_DECLARATIONS = [
+    NggModalComponent,
+    NggModalHeaderComponent,
```
Maximilian-R, line 8: Header, Body & Footer is just internal directives for the Modal Component right? If so, skip exporting them

splashdust:
@SEBNiklas Have you considered using the [dialog](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) element? That could also, at least theoretically, solve the backdrop issue with IBP, using the [::backdrop](https://developer.mozilla.org/en-US/docs/Web/CSS/::backdrop) pseudo element.

Maximilian-R (changes requested):

libs/angular/src/lib/modal/modal.component.ts
```diff
@@ -0,0 +1,130 @@
...
+        this.closeModal(event);
+    }
+
+    public handleBackdropClick(event: MouseEvent) {
+        if (event.target == this.backdropRef?.nativeElement)
+            this.closeModal(event);
+    }
+    
+    public handleDismiss(event: MouseEvent) {
+        if (this.onDismiss.observers.length > 0)
```
Maximilian-R, line 54: is there a reason you only emit if there is an observer?

vsjolander:
Closes #313

vsjolander:
@SEBNiklas do you have time to work on this PR or do you need any help from us?

libs/angular/src/lib/modal/modal.component.ts
SEBNiklas: For `onDismiss` and `onConfirm` no, but for `onClose` yes. For `onClose` its to be able to intercept the closing of the modal outside of the component, where you can have some logic determine if you want to close the modal or not by updating the `isOpen` property. For example to show a confirmation dialog (Are you sure you want to close? Yes or No) when you close (e.g. press the close/cross button) the modal.

But for `onDismiss` and `onConfirm` its not necessary to only emit if there are any observers, but I guess you would only want to emit if you have any observers? But perhaps that check is already done in emit?

libs/angular/src/lib/modal/modal.module.ts
SEBNiklas: Yes, that's right. I will change that. 

As a side note, a future improvement would perhaps be to allow other content then just a text/string in the header and also have other content in the footer then just the two default buttons. I'm mentioning this because its something we have seen a need for in our project.

libs/angular/src/lib/modal/modal.service.ts
SEBNiklas: Yes that class is missing, I forgot about that. In our project we placed this class in the styles.scss (toplevel/root scss-file), since we couldn't place it in the scss-file for the component because of component encapsulation.

Anyway, it sounds like a good idea to use the same library in this case as well. Will change that.

libs/angular/src/lib/modal/modal.component.ts
Maximilian-R: Gotcha! :) 

Yes, if the emitter doesn't have any registered observers it will simply not call any. So there is no need to check the amount of observers here, expect for `onClose` then 👍

SEBNiklas:
> @SEBNiklas Have you considered using the [dialog](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) element? That could also, at least theoretically, solve the backdrop issue with IBP, using the [::backdrop](https://developer.mozilla.org/en-US/docs/Web/CSS/::backdrop) pseudo element.

Didn't know about that. The support in browsers seems to have been added very recently though, at least for mobile browsers. I don't know if that's an issue in Green? What browsers does green aim to support and does it only support most recent version of said browsers?

In this case perhaps we could use ::backdrop, but isn't still pretty likely that other classes in the shells will affect MFEs using green? More so than MFEs using bootstrap, since the shells are using bootstrap. Do you know if the shells have any plans to start using Green or use scope (same as green) to not affect the MFEs? Perhaps they are already using scope and this is just a single issue?

vsjolander:
> We have planned to work on this PR next week. We hope to complete this PR but depending on how much times it will take, we may need your help.

We have Green Coaching meetings on Mondays 13:00 where we would love to help you with these kinds of things. You can book a slot in Teams under Green Design System > Book a session, there's a tab in this channel for Green Coaching booking.

If you rather book a separate meeting that's OK too :)

SEBNiklas:
@vsjolander I think I have finished all the work on this PR except for the issue with backdrop, which I'm not sure should be solved in green? but rather in IBP? We have noticed that alert and badge also gets affected by IBP.

Also I noticed that the tests fail, but it seems to be the test for the dropdown component.

Maximilian-R:
Cool :) @vsjolander looks like the tests are failing because of the _popover change, will you have a look?

Maximilian-R (approved):

splashdust:
> > @SEBNiklas Have you considered using the [dialog](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) element? That could also, at least theoretically, solve the backdrop issue with IBP, using the [::backdrop](https://developer.mozilla.org/en-US/docs/Web/CSS/::backdrop) pseudo element.
> 
> Didn't know about that. The support in browsers seems to have been added very recently though, at least for mobile browsers. I don't know if that's an issue in Green? What browsers does green aim to support and does it only support most recent version of said browsers?
> 
> In this case perhaps we could use ::backdrop, but isn't still pretty likely that other classes in the shells will affect MFEs using green? More so than MFEs using bootstrap, since the shells are using bootstrap. Do you know if the shells have any plans to start using Green or use scope (same as green) to not affect the MFEs? Perhaps they are already using scope and this is just a single issue?

Yeah, this may need further investigation, both in terms of browser support (although I think we should be OK with a polyfill) and shell compatibility. But let's drop it for the scope of this PR :)

splashdust (approved):

vsjolander (approved):

PR #624, opened 2022-10-26
style: 💄 removing red from charts

vsjolander:
Red was seen as a negative color by users so it should not be used in charts


splashdust (approved):

sebopensource:

JohanObrink (approved):

PR #632, opened 2022-11-02
feat: 🎸 New focus outline

vsjolander:
This is a new design for the focus outline to be a 2px black border instead of boxshadow which is not very accessible in it's contrast.

This is WIP and the PR is used to share a storybook right now.

### DIFF
--- a/libs/chlorophyll/scss/common/_focus.scss
+++ b/libs/chlorophyll/scss/common/_focus.scss
@@ -2,7 +2,7 @@
 @use 'functions';
 @use 'sass:map';
 
-@mixin add-outline($color: tokens.$black) {
+@mixin add-outline($color: tokens.$grey-1000, $offset: 2px) {
   outline: 2px solid $color;
   outline-offset: 2px;
 }
--- a/libs/chlorophyll/scss/common/_focus.scss
+++ b/libs/chlorophyll/scss/common/_focus.scss
@@ -2,12 +2,20 @@
 @use 'functions';
 @use 'sass:map';
 
-@mixin add-outline($color: tokens.$grey-1000, $offset: 2px) {
-  outline: 2px solid $color;
+@mixin add-outline($color: tokens.$grey-1000, $offset: .125rem) {
+  outline: .125rem solid $color;
   outline-offset: $offset;
 }
 
-@mixin add-focus($suffix: 'visible', $target: '', $offset: 2px) {
+/**
+  Styling for focus state
+
+  @param string $suffix - a suffix for focus so mixin can be used for focus, focus-visible and focus-within
+  @param string $target - selector for a target that is a child of seletor where you place mixin
+  @param length $offset - param used to set outline-offset
+ */
+
+@mixin add-focus($suffix: 'visible', $target: '', $offset: .125rem) {
   @if $suffix != '' {
     $suffix: '-' + $suffix;
   }
@@ -41,7 +49,7 @@
     @include add-outline(tokens.$grey-1000, $offset);
 
     &:active {
-      outline-offset: 0px;
+      outline-offset: 0;
     }
   }
 }
@@ -51,8 +59,8 @@
     $color: transparentize($color, 0.35);
   }
   &:focus {
-    outline: $color solid 2px;
-    outline-offset: 2px;
+    outline: $color solid .125rem;
+    outline-offset: .125rem;
   }
 
   &:focus:not(:focus-visible) {
@@ -60,8 +68,8 @@
   }
 
   &:focus-visible {
-    outline: $color solid 2px;
-    outline-offset: 2px;
+    outline: $color solid .125rem;
+    outline-offset: .125rem;
   }
 }
 
--- a/libs/chlorophyll/scss/common/_focus.scss
+++ b/libs/chlorophyll/scss/common/_focus.scss
@@ -40,8 +40,7 @@
 
   @if $suffix != '' {
     &:focus:not(:focus-visible) {
-      box-shadow: none;
-      outline: 0;
+      outline: none;
     }
   }
 
@@ -124,13 +123,15 @@
   @include add-active-state($active, $fadeActiveHover);
 }
 
-/** add background color, color and border-color to element when it has focus-visible i.e. tab focus */
+/**
+  add background color, color and border-color to element when it has focus-visible i.e. tab focus
+*/
 @mixin add-active-focus-state($color: tokens.$black) {
   &:focus-visible {
-    //@include functions.desired-color-combination(tokens.$white, $color, 20%, 3);
-    /*border-color: nth(
+    @include functions.desired-color-combination(tokens.$white, $color, 20%, 3);
+    border-color: nth(
       functions.desired-color-combination(tokens.$white, $color, 20%, 3),
       1
-    ) !important;*/
+    ) !important;
   }
 }
--- a/libs/chlorophyll/scss/common/_focus.scss
+++ b/libs/chlorophyll/scss/common/_focus.scss
@@ -2,25 +2,33 @@
 @use 'functions';
 @use 'sass:map';
 
+@mixin add-outline($color: tokens.$black) {
+  outline: 2px solid $color;
+  outline-offset: 2px;
+}
+
 @mixin add-focus($suffix: 'visible', $target: '') {
   @if $suffix != '' {
     $suffix: '-' + $suffix;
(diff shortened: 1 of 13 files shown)


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybooks we've created for it: <ul><li><a target="_blank" href="https://sebgroup.github.io/green/pr-632/chlorophyll/">chlorophyll</a></li><li><a target="_blank" href="https://sebgroup.github.io/green/pr-632/react/">react</a></li></ul>

libs/chlorophyll/scss/common/_focus.scss
```diff
@@ -2,25 +2,33 @@
 @use 'functions';
 @use 'sass:map';
 
-@mixin add-focus($suffix: 'visible', $target: '') {
+@mixin add-outline($color: tokens.$grey-1000, $offset: 2px) {
+  outline: 2px solid $color;
```
toplizard, line 6: can we use rem, so that the frame will scale when zooming

libs/chlorophyll/scss/common/_focus.scss
```diff
@@ -2,25 +2,33 @@
...
+  outline-offset: $offset;
+}
+
+@mixin add-focus($suffix: 'visible', $target: '', $offset: 2px) {
```
toplizard, line 10: Can you add documentation to explain what the properties are

libs/chlorophyll/scss/common/_focus.scss
```diff
@@ -30,10 +38,11 @@
...
-    border-color: tokens.get(tokens.$blue, 0) !important;
-    box-shadow: 0 0 0.25em 0.0625em tokens.get(tokens.$blue, 2);
-    outline-color: transparent;
-    outline-style: solid;
+    @include add-outline(tokens.$grey-1000, $offset);
+
+    &:active {
+      outline-offset: 0px;
+    }
   }
```
toplizard, line 46: we need a :focus:not(:focus-visible) for mouse users, i guess it should be no outline in that case

libs/chlorophyll/scss/common/_focus.scss
```diff
@@ -109,12 +117,12 @@
...
-@mixin add-active-focus-state($color: tokens.get(tokens.$blue, 2)) {
+@mixin add-active-focus-state($color: tokens.$black) {
   &:focus-visible {
-    @include functions.desired-color-combination(tokens.$white, $color, 20%, 3);
-    border-color: nth(
+    //@include functions.desired-color-combination(tokens.$white, $color, 20%, 3);
+    /*border-color: nth(
       functions.desired-color-combination(tokens.$white, $color, 20%, 3),
       1
-    ) !important;
```
toplizard, line 118: Do we need important?

libs/chlorophyll/scss/components/badge/_mixins.scss
```diff
@@ -31,16 +31,15 @@ $variants: (
...
 $color: var(--color);
 $background: var(--background);
 $close-bg-hover: hsla(var(--sg-hsl-black), 0.2);
 $close-bg-active: hsla(var(--sg-hsl-black), 0.3);
 
 $close-height: 1.5rem;
 $close-width: $close-height;
-
-$close-icon-thickness: 3px;
+$close-icon-thickness: 2px;
```
toplizard, line 43: can we use rem here?

libs/chlorophyll/scss/components/form/checkbox/_mixins.scss
```diff
@@ -44,7 +44,6 @@ $validated-selector: '.was-validated';
     background-color: $background;
     flex: 0 0 auto;
     height: 1.25rem;
-    transition: 300ms;
```
toplizard, line 47: I think that there are transition variables

libs/chlorophyll/scss/components/link/_index.scss
```diff
@@ -1,19 +1,17 @@
 @use 'mixins';
 @use 'sass:map';
 
-
```
toplizard, line 4: i think it would be nice if we where to use scoped classes eg. "seb-green-link", that whay it's less prone to leak styling, also it is easier to understad where the styling is coming from when inspecting elements in the browser

libs/chlorophyll/scss/components/link/_mixins.scss
```diff
@@ -28,12 +28,17 @@ $variants: (
...
   text-decoration: none;
   border: none;
 
+  p > & {
+    border-radius: 2px;
+    outline-offset: 3px;
+  }
+
   &:hover,
   &:focus {
```
toplizard, line 43: pls always use rem

libs/chlorophyll/scss/components/link/_index.scss
vsjolander: Green should always be scoped to .use-green to prevent styling, that's the idea right now.

Having tailwind style classes might cause developers to "implemented their own component" with utility classes and future changes to design system could cause breaks. Having those cases as specific components is better to get the feedback loop in place.

splashdust: Just to add a bit of background here, it was decided in discussions early on that one of the goals of Chlorophyll should be to make markup look correct "by default".

I.E. the following snippet of markup should render a correct SEB-styled document, provided you load the Chlorophyll CSS:
```html
<h1>Hello world</h1>
<p>Lorem ipsum <a href="/url">dolor</a> sit amet</p>
<button>Do something!</button>
```

In other words, the lack of scoped classes is "by design" :)

But I think this is a discussion we should keep having. It seems likely that this approach will keep causing problems in all sorts of scenarios in the future. Perhaps it would be good enough that the React and Angular components looks correct by default, and that Chlorophyll requires prefixed classes for everything.

libs/chlorophyll/scss/common/_focus.scss
vsjolander: There is on line 34 :)

libs/chlorophyll/scss/components/link/_index.scss
vsjolander: I agree, we should keep having this discussion! 

It will cause problems in all sorts of scenarios but I think that could be a good thing. 

If a such a layout/styling problems occurs: 
1. Either the markup is faulty and needs to be corrected or 
3. Green needs to get that contribution/feedback in order to update styles. And nothing gets fixed faster than things that doesn't **look** good 😄

libs/chlorophyll/scss/common/_focus.scss
```diff
@@ -42,25 +58,24 @@
...
 
   &:focus:not(:focus-visible) {
     outline: 0;
   }
 
   &:focus-visible {
-    outline: $color solid 2px;
-    outline-offset: 2px;
+    outline: $color solid .125rem;
+    outline-offset: .125rem;
```
patrikfogelstrand, line 71: I don't think this will work properly for Safari, we've seen that behavior when trying `focus-visible` stuff there.

EldRoos: Patrik, how long ago was this?

> This is great news! Safari has long been the missing link here, but with the up-coming Safari 15.4, :focus-visible will have full support across modern browsers.
https://www.tempertemper.net/blog/safari-focus-visible-and-accessibility

Safari have version 16.3 right now. 15.4 was released March 15, 2022.

patrikfogelstrand: It was earlier this month, I think we tested both Safari 15 and 16. We ended up using something like this:

```scss
@mixin focus-outline($color: $black, $size: 0.125rem) {
  &:focus {
    outline: $size solid $color;
    outline-offset: $size;
  }
  &:focus:not(:focus-visible) {
    box-shadow: none;
    outline: 0;
  }
}

```

patrikfogelstrand:
Make sure to check if the focus outline styles with `:focus` and `:focus-visible` behaves as expect when setting the focus programmatically, using default JS `element.focus()` or a library like [focustrap](https://github.com/focus-trap/focus-trap). 

We've had issues finding a good focus mixin that works for all uses cases (mouse, touch, keyboard and programmatically), where it was either Safari or setting focus programmatically that caused issues.

splashdust:
Closing this PR now, since it has gotten very out of date with the main branch. If these changes are still relevant it would be better to dig them out into a new PR.

PR #636, opened 2022-11-03
Make Extract SSR compatible

### LINKED ISSUE #631: React components not usable with server side render
Currently it is not possible to use the React components in a server side rendered context. The problem stems from the use of a library called `Inputmask` in Extract.

Inputmask expect the global object `self` to be available, but it only exist in the browser and therefore it fails with a `ReferenceError` on the server.

The same problem will likely exist in Angular as well.

splashdust:
Fixes #631


sebopensource:

PR #641, opened 2022-11-09
fix: 🐛 FormControl touched property update/reset fix

nerjok:
FormControl touched property should be updated after blur event and should be untouched after control reset() is called

BREAKING CHANGE: 🧨 -

✅ Closes: -

After formControl.reset() is called status should be reset and later marked as touched on blur event.

```
True if the control is marked as touched.

A control is marked touched once the user has triggered a blur event on it.
```
https://angular.io/api/forms/AbstractControl#touched

### DIFF
--- a/libs/angular/src/lib/dropdown/dropdown.component.ts
+++ b/libs/angular/src/lib/dropdown/dropdown.component.ts
@@ -171,10 +171,12 @@ export class NggDropdownComponent
       searchFilter: this.searchFilter,
       compareWith: this.compareWith,
       onTouched: () => {
-        this.onTouchedFn?.()
         this.touched.emit(true)
         this.cd.markForCheck()
       },
+      onBlur: () => {
+        this.onTouchedFn?.()
+      }
     }
   }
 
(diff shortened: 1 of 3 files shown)


Maximilian-R:
Hi, can you describe in what scenario you need this change? :) 

I would argue that the rule "A control is marked touched once the user has triggered a blur event on it." is not applicable on the dropdown since there is a false positive when opening the dropdown (clicking the button to open the list causes a blur event). Only when focus leaves the "Dropdown" as a whole component onTouched should be called

nerjok:
> ld argue that the rule `A control is marked touched once the user has triggered a blur event on it.` targets the control as "a whole" and not the dropdown button alone. There is a false positive when opening the dropdown and moving into the dropdown list with arrow keys (a blur event will be triggered on the dropdown button). Only when focus leaves the dropdown as a whole component onTouched should be called
> 
> This is what the current implementation of the onTouched

thanks overviewed codebase a bit more, will remove this PR. created issue with description, for now solved by manually triggering ngAfterViewInit after reset, will overview other options.



But the issue i try solve --- selecting option in dropdown

    <ngg-dropdown
      formControlName="country"

selecting option manualy -> reseting control  .get('country')?.reset()

and here touched/untouched state is no longer changing.

PR #644, opened 2022-11-10
fix(stepper): only emit onChange on user events

Maximilian-R:
onChange should not be emitted when the stepper is created - and not when setting the value from a parent component (eg: when the component is used like a controlled input)


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we've created for it: <ul><li><a target="_blank" href="https://sebgroup.github.io/green/pr-644/react/">react</a></li></ul>

splashdust (approved):

PR #646, opened 2022-11-15
Syncing design from Figma

vsjolander:
Many components are not in sync with the design in Figma.

In this PR I've synced:

- Alerts
- Checkboxes
- Labels
- Information label

### DIFF
--- a/apps/react-lib-dev/src/app/app.tsx
+++ b/apps/react-lib-dev/src/app/app.tsx
@@ -32,6 +32,7 @@ export function App() {
 
   return (
     <>
+    <div className="use-green">
       <Navbar title="Green React Dev" />
       <div className="container">
         <div className="row">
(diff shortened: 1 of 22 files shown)


Maximilian-R (commented):
Awesome changes! :) 

Perhaps mark this as a breaking change if its not that already, some existing form layouts might break. 

Found one issue with the validation line on "group", it takes 100% of parent even if fields does not. Might be difficult to solve when its placed on the info span below, or let the group items grow to full width?
https://sebgroup.github.io/green/pr-646/chlorophyll/?path=/docs/components-form-concepts-validation--page

splashdust (approved):

vsjolander:
Good points!

Fixed the sizing of inputs in groups so it fills up together with the validation line.
Also marked the breaking change, not sure if I did it correct though 😸

PR #649, opened 2022-11-16
Component: Accordion

vsjolander:
This is draft PR that will be used during a mob programming session.


vsjolander:
Accordion was contributed by @SEBNori

PR #660, opened 2022-11-23
feat: 🎸 dropdown add formControl reset functionality

nerjok:
reset field form control when user programatically formControl.reset() field and update touched state after user is leaving field

BREAKING CHANGE: - -

✅ Closes: 643

[issue](https://github.com/sebgroup/green/issues/643)

### DIFF
--- a/libs/angular/src/lib/dropdown/dropdown.component.ts
+++ b/libs/angular/src/lib/dropdown/dropdown.component.ts
@@ -2,16 +2,36 @@ import {
   AfterViewInit,
   ChangeDetectionStrategy,
   ChangeDetectorRef,
-  Component, ContentChild, ElementRef, EventEmitter, Input,
+  Component,
+  ContentChild,
+  ElementRef,
+  EventEmitter,
+  Inject,
+  Injector,
+  Input,
   OnChanges,
   OnDestroy,
-  Output, SimpleChanges, ViewChild
+  Output,
+  SimpleChanges,
+  ViewChild,
 } from '@angular/core'
-import { ControlValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms'
 import {
-  AbstractDropdown, CompareWith, createDropdown, DropdownArgs, DropdownHandler,
+  ControlValueAccessor,
+  NgControl,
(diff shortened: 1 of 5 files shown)


libs/angular/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sebgroup/green-angular",
-  "version": "1.0.0-beta.38",
+  "version": "1.0.0-beta.39",
```
vsjolander, line 3: No need to bump version, is handled when publishing.

libs/angular/src/lib/dropdown/dropdown.component.ts
```diff
@@ -125,6 +135,14 @@ export class NggDropdownComponent
 
   writeValue(value: any): void {
     this.value = value
+    if (
+      value === null &&
+      this.dropdown?.isTouched &&
+      this.control?.control?.untouched
+    ) {
+      this.setHandler()
```
vsjolander, line 143: Do we need to create and set the handler again or can we update the current one? I'm out on a limb here :)

vsjolander (commented):
Looks good, had some minor questions. Would like Max and Joacim to review since they are more into Angular! 👍

libs/angular/src/lib/dropdown/dropdown.component.ts
nerjok: i was thinking about that also, if wouldnt be any side effects to change only `handler.dropdown.isTouched` property  on `formControl.reset()`

Any thoughts from others?

Maximilian-R: Yes that sounds like a good idea, it would not have any side effects - but make sure to change the condition on line 160 in green\libs\extract\src\lib\dropdown\dropdown.ts to only trigger when the new value is true.

Also don't modify the isTocuhed property directly, create a new reducer in extracts and make it callable. Let me know if you need any clarification or help :)

nerjok: thanks, updated PR with suggestions, i hope didnt miss anything now.

Maximilian-R (approved):

PR #669, opened 2022-11-30
fix: 🐛 radio alignment and colors

mario-subo:
change borders to box-shadow to fix alignment and clean up scss readibility


vsjolander (approved):
@mario-subo  did some small changes to how the dot is styled. Other than that I think it looks great 👍

PR #674, opened 2022-12-05
Fix/checkbox react

vsjolander:


libs/chlorophyll/scss/components/alert-ribbon/_mixins.scss
```diff
@@ -102,12 +102,12 @@ $link-outline-color: hsla( var(--sg-intent-neutral-hsl-contrast), link.$outline-
       display: inline;
       font-size: 1rem;
       font-weight: 500;
-      @include common.line-height(3);
```
isapet-4, line 105: Should we not use common.line-height any longer? 
If not, isn't it better if we remove it? Otherwise, we should use it here as we did before, but with correct rem size.

libs/chlorophyll/scss/components/form/checkbox/_mixins.scss
```diff
@@ -53,33 +56,14 @@ $invalid-color: tokens.$intent-danger-background;
     opacity: 1;
   }
 
-  /* Checked state */
+  // Checked state
   #{$selector} #{$checkbox-selector}:checked ~ i {
     background-color: var(--gds-comp-checkbox-container-color-selected);
   }
 
-  /* Valid state */
```
isapet-4, line 61: Not sure how you want to do it, but couldn't we remove commented code? It's still in the git history if we want to revert it later on.

libs/chlorophyll/scss/components/alert-ribbon/_mixins.scss
vsjolander: For some reason there was 2 line-height scales and I wasn't sure which rule was written with which scale. So I decided to remove the use in this mixing and explicitly write the value so I don't get it wrong.

Then maybe if there is a need to use the scale - which I don't really see at this point - we can go back and use it... 

But yea, maybe remove it since we don't use it..

libs/chlorophyll/scss/components/form/checkbox/_mixins.scss
vsjolander: Yes! It is removed 👍

isapet-4:
Looks good, I do not have sufficient access to approve the PR though.

PR #680, opened 2022-12-08
React: Accordion

vsjolander:
First version of an Accordion component in React


splashdust (approved):
Looks good!

libs/react/src/lib/accordion/accordion-item.tsx
```diff
@@ -0,0 +1,63 @@
...
+          <span>{label}</span>
+          { subLabel && <span>{subLabel}</span> }
+          <svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
+            <path
+              d="M18.8095 9.22817L18.1907 8.60942C18.0438 8.46255 17.8063 8.46255 17.6595 8.60942L12.0001 14.2563L6.34072 8.60942C6.19385 8.46255 5.95635 8.46255 5.80947 8.60942L5.19072 9.22817C5.04385 9.37505 5.04385 9.61255 5.19072 9.75942L11.7345 16.3032C11.8813 16.45 12.1188 16.45 12.2657 16.3032L18.8095 9.75942C18.9563 9.61255 18.9563 9.37505 18.8095 9.22817Z"
+              fill="#333333"/>
+          </svg>
+        </button>
+      </div>
+      <div role="region" aria-labelledby={`accordion-${index}-${uuid}`} style={{height: isOpen ? height : 0}}>
```
Maximilian-R, line 54: I think aria-hidden & hidden attribute should be applied on the item when it's closed to keep screen readers and keyboard focus from content within while its closed. @splashdust didnt you fix something like this in public web accordion recently? What solution did you apply there?

vsjolander: Yes hidden is the way to go, forgot to add this.

Tried with aria-hidden, but you could still tab into it.

splashdust: @Maximilian-R Good catch!
@vsjolander You can use the `inert` attribute in combination with aria-hidden to make the element both undiscoverable and uninteractable, but still rendered. Should solve the height animation issue.
https://developer.chrome.com/articles/inert/

PR #704, opened 2023-01-13
feat: 🎸 Angular modal focus functionality

SEBNiklas:
Added functionality to trap focus inside modal (Angular)


vsjolander (approved):
Looks good! 

Would need some test for this functionality but that the case for many components right now...

PR #715, opened 2023-01-25
feat: 🎸 add new large button size for jumbotrons/heroes

sjkayle:
Preview:

image

Use-case:

image

### DIFF
--- a/libs/angular/src/lib/button/button.component.ts
+++ b/libs/angular/src/lib/button/button.component.ts
@@ -4,7 +4,7 @@ import {
   HostBinding,
   Input
 } from '@angular/core'
-import { ButtonType, ButtonVariant } from '@sebgroup/extract';
+import { ButtonSize, ButtonVariant } from '@sebgroup/extract';
 
 @Component({
   // eslint-disable-next-line @angular-eslint/component-selector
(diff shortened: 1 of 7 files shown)


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybooks we've created for it: <ul><li><a target="_blank" href="https://sebgroup.github.io/green/pr-715/chlorophyll/">chlorophyll</a></li><li><a target="_blank" href="https://sebgroup.github.io/green/pr-715/react/">react</a></li><li><a target="_blank" href="https://sebgroup.github.io/green/pr-715/angular/">angular</a></li></ul>

mario-subo:
it seems like your change introduces a new class name `"large"` but it does not seem to remove the `"size-lg"` class which applies the same styles. I noticed that there is also a modal component which does a similar thing so maybe there needs to be a better standard for class names regarding sizing for chlorophyll before we proceed with this change (or perhaps as part of this PR?. @vsjolander any thoughts on this?

vsjolander (approved):
Awesome!

vsjolander:
> it seems like your change introduces a new class name `"large"` but it does not seem to remove the `"size-lg"` class which applies the same styles. I noticed that there is also a modal component which does a similar thing so maybe there needs to be a better standard for class names regarding sizing for chlorophyll before we proceed with this change (or perhaps as part of this PR?. @vsjolander any thoughts on this?

Oh missed this before reviewing and merging... Yes I agree we should set a standard..

To be honest I hadn't seen the `size-lg` class before 🙈 Maybe this could be quite nice to actually have a `[property]-[value]` standard for these things. We can tie it to what the property name is in Figma and what the prop is named in React or Angular.

@splashdust @Maximilian-R do you have any input on this?

Since this has to be refactored everywhere I merge these changes since they are good enough for now! But would like to set a standard for sure!

PR #723, opened 2023-01-31
Feature/select

vsjolander:


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybooks we've created for it: <ul><li><a target="_blank" href="https://sebgroup.github.io/green/pr-723/chlorophyll/">chlorophyll</a></li><li><a target="_blank" href="https://sebgroup.github.io/green/pr-723/react/">react</a></li></ul>

splashdust:
@vsjolander Should `Select` and `Dropdown` really be separate components? Semantically, functionally and visually they are basically the same, so perhaps *native select* should just be the default render method for the existing `Dropdown` component? (or an optional render method to avoid breaking changes perhaps)

Or maybe it makes sense to have both `<Select>` and `<Dropdown>` in React, but describe them as variations of the same component in the docs (i.e still a single entry in the Components menu).

vsjolander:
We (@splashdust, @isapet-4 @daneriseb) had a meeting discussing this component.

We reach the conclusion that Select is a good name for a new module which should contain the components "Select" and "MultiSelect" that should be using the native select html elements in them. Form elements should in most cases be one of these components.

The "Dropdown" component should probably be deprecated as is since it's not accessible. Maybe it should be called "RichSelect" and be a part of the Select module.

We also need to clarify the difference between a "Fold-out" and a "Dropdown". What is a navigational component and what is a form component?

PR #724, opened 2023-02-01
fix(slide-toggle): 🐛 move transition to hover

patrikfogelstrand:
The transition being set directly in the mixin caused some awkward visual effects when using e.g. `visibility: hidden` on a parent element. Moving the transition styling to appear only on hover effect (which is the intended use of it), keeps the visual transition effect when needed, but does not give awkward visual effects when using `visibility:hidden`.

To reproduce:

- go to [storybook: slide-toggle](https://sebgroup.github.io/green/latest/chlorophyll/?path=/docs/components-slide-toggle--slide-toggle)
- open dev tools and apply `visibility:hidden;` on a parent element to the slide-toggle button
- watch how it disappears from the view with an awkward transition

The new code will only apply the transition when hovering and you can toggle the visibility of a parent element as much as you want.

### DIFF
--- a/libs/chlorophyll/scss/components/slide-toggle/_mixins.scss
+++ b/libs/chlorophyll/scss/components/slide-toggle/_mixins.scss
@@ -10,6 +10,7 @@ $_track_height: 1.5rem;
 $_track_width: 2.75rem;
 $_track_height_mobile: 2rem;
 $_track_width_mobile: 3.5rem;
+$_track_transition_time: 150ms;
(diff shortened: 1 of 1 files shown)


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we've created for it: <ul><li><a target="_blank" href="https://sebgroup.github.io/green/pr-724/chlorophyll/">chlorophyll</a></li></ul>

splashdust (approved):

Issue #733, opened 2023-02-14
Align checkboxes to the top when long lables


vsjolander:
<img width="376" alt="Screenshot 2023-02-14 at 16 27 04" src="https://user-images.githubusercontent.com/11420341/218782199-762eacaa-d9dd-4c04-8ffa-b8fb51ba04a4.png">

astrit:
If you would `align-self: start;` the checkmark it self and perhaps add some margin on top for the visual parity, should fix it. 
Just playing around on this fork for future reference ✌️

Issue #749, opened 2023-02-24
Checkboxes are vertically aligned in the middle of the label


daneriseb:
It doesn't look good when the label text is wrapped.

![image](https://user-images.githubusercontent.com/113447576/221205007-b78749b0-ec92-49e8-8d30-f041a30c4fa1.png)

daneriseb:
@vsjolander Should this be closed? It seems to be fixed by https://github.com/sebgroup/green/pull/756

PR #756, opened 2023-03-02
Feat/checkbox horizontal & alignment issue

vsjolander:
Closes: #733 #749


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybooks we've created for it: <ul><li><a target="_blank" href="https://sebgroup.github.io/green/pr-756/chlorophyll/">chlorophyll</a></li><li><a target="_blank" href="https://sebgroup.github.io/green/pr-756/react/">react</a></li></ul>

Maximilian-R (approved):
I just updated the html for all the form components in an angular project using green to align with green 1.0. And with this commit some of them must be updated again to add those wrapper. Proves that we really should implement all the components in Angular as well - never trust that the chlorophyll html will be immutable just because its correct semantics!

splashdust:
@Maximilian-R What in particular causes the incompatibility? Maybe this change needs to either keep the old style along with the new one, for backwards compatibility, or be tagged as a breaking change.

Maximilian-R:
> @Maximilian-R What in particular causes the incompatibility? Maybe this change needs to either keep the old style along with the new one, for backwards compatibility, or be tagged as a breaking change.

Haven't tried if it acually breaks something, but they fact that it solves an issue that requires angular projects to update their html to benefit of the same change is crazy 😄

vsjolander:
I've kept all old selectors and don't think I've fiddled with them just to be backwards compatible.

From what I can see in the [chlorophyll storybook](https://sebgroup.github.io/green/pr-756/chlorophyll/) (it has only the "old" syntax with .form-groups for form items, check the concepts pages) it looks like things are holding up! Not sure if I should update Chlorophyll with this new .gds-form-item syntax.. what do you think?

Chlorophyll feels like kind of like a bootstrap child, simple markup with simple "component classes" and modifiers. But once you get in to more advanced styling and component you want to wrap them and more control over design and functionality and I think that's where we are headed..?

What you are saying Max, changes to chlorophyll doesn't cause markup in apps to update. And I get the feeling from the teams channel that developers want a wrapped component that comes with "batteries included". So I think it's time Green got some more angular components :)

Maximilian-R:
> I totally get what you are saying @Maximilian-R, changes to chlorophyll doesn't cause markup in apps to update. And I get the feeling from the teams channel that developers want a wrapped component that comes with "batteries included". So I think it's time Green got some more angular components :)

Give the people what they want! 🧙 Some component wrappers would make "good-first-issue" if anyone is interested to help out with Green 😄

splashdust:
> From what I can see in the [chlorophyll storybook](https://sebgroup.github.io/green/pr-756/chlorophyll/) (it has only the "old" syntax with .form-groups for form items, check the concepts pages) it looks like things are holding up!

Ah, yes, it looks unaffected!

> Not sure if I should update Chlorophyll with this new .gds-form-item syntax.. what do you think?

I'm not sure if I've understood the reason for the new class. Does it deprecate `.form-group` , or does it have a different purpose? 

I agree on the wrapper components for Angular!

PR #758, opened 2023-03-03
fix(chlorophyll): form-group margin in mobile

Maximilian-R:


libs/chlorophyll/scss/components/form/common/_mixins.scss
```diff
@@ -110,10 +110,7 @@
   position: relative;
   width: 100%;
   z-index: 0;
-
-  @include common.media-breakpoint-up('sm') {
```
Maximilian-R, line 114: I cannot see any reason why this would not be applied to small screens as well. There is a another common rule `form, .form { @include common.margin-bottom(6); } ` which applies the same margin to any screen size. I want to align the behavior so that it works the same on mobile screens when the .form-group is not a direct child

sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we've created for it: <ul><li><a target="_blank" href="https://sebgroup.github.io/green/pr-758/chlorophyll/">chlorophyll</a></li></ul>

libs/chlorophyll/scss/components/form/common/_mixins.scss
```diff
@@ -110,10 +110,7 @@
...
-    @include common.margin-bottom(6);
-  }
+  @include common.margin-bottom(6);
```
Maximilian-R, line 113: I cannot see any reason why this would not be applied to small screens as well. There is a another common rule form, .form { @include common.margin-bottom(6); }  which applies the same margin to any screen size. I want to align the behavior so that it works the same on mobile screens when the .form-group is not a direct child

Maximilian-R: Questions is, can you think of any usecase where this might deviate? All our examples in storybook has the form groups as direct children of form anyways :) @vsjolander @splashdust

splashdust: I agree, this seems like an unexpected inconsistency. I cannot think of a use case where this would be desirable.

splashdust (approved):

libs/chlorophyll/scss/components/form/common/_mixins.scss
vsjolander: I agree as well. I think this remains from when Green Code and Green Design wasn't as synced as it is now :)

vsjolander (approved):

Issue #725, opened 2023-02-01
Improve developer experience on Windows machine


vsjolander:
It's not optimal right now.

Seems like we still rely on node-gyp witch is a pain to install on Windows machines since it needs SDKs and other dependancies making the installation process far to long.

This needs to be optimized....

patrikfogelstrand:
I think it's really important to differentiate between dev experience on a SEB Windows machine and any Windows machine. 

For any Windows:
- node-gyp is a (sub)dependency of the repo, which needs both [Python and a Windows SDK](https://github.com/nodejs/node-gyp#on-windows) installed. This is just a real pain, as a developer for a Node based application, I would prefer not having to install Python, Windows SDK, Visual Studio (full fledged), C++ tools, etc. 

Additional stuff for SEB Windows:
- different git configs needed for on-prem vs. cloud github ([solution](https://stackoverflow.com/a/43884702/12416767) but requires a lot of config and understanding of SEB networking)
- yarn seems pretty much impossible to use, atleast without any major config

splashdust:
Many improvements and upgrades have been made since this issue was opened. It's still far from ideal, but this issue is very open ended, so I'm closing it now, and new and more specific issues can be opened as needed.

PR #776, opened 2023-03-10
chore(deps): remove `node-sass` from dependencies

splashdust:
`node-sass` is replaced by `sass`, so we can remove this dependency. One step towards improved dev experience! (#725)

The only thing that still depends on `node-gyp` is `@angular/cli@13.3.3`


splashdust:
Got some seemingly unrelated test failures in React/Dropdown. It turned out to be a simple timing issue in the tests, but found some other issues with the Datepicker that gets corrected in this PR too.

splashdust:
> It turned out to be a simple timing issue in the tests

Or so I though... There might be a deeper issue at play here. Continuing to investigate.

splashdust:
Turns out that the error message in the test logs, `When testing, code that causes React state updates should be wrapped into act(...)` means that code that causes React state updates should be wrapped in `act` and not just waited for arbitrary milliseconds. Who could have guessed!

PR #785, opened 2023-03-15
docs: ✏️ replacing self closing i-tag with open and close tags

vsjolander:


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we've created for it: <ul><li><a target="_blank" href="https://sebgroup.github.io/green/pr-785/chlorophyll/">chlorophyll</a></li></ul>

splashdust (approved):
Nice! We should probably try to get rid of these empty `i` elements altogether in the future. If I'm not mistaken, they are basically just styling crutches for the checkbox tick marks.

astrit (approved):
Looks good to me and I do agree with @splashdust in future perhaps we should move away from using the `i` tag.

PR #801, opened 2023-03-20
Chore/upgrade deps

splashdust:
# Changes
The PR upgrades NX, Storybook, React, Angular and a bunch of other dependencies to the latest version.

## Reviewers:
Most of the changes are generated by `nx migrate`. The best way to review this PR is probably to clone it locally and verify that the storybooks and tests still work on your machine.


splashdust:
If anyone wonders about the `chore(chlorophyll): release version 1.4.0` commits; they were accidental during testing when upgrading the `semver` package. We'll just filter those out when squashing and merging!

astrit (approved):
Looks good to me, great job @splashdust   🙌

Maximilian-R:
Cannot npm install from a clean state, unable to resolve zone.js version:

image

splashdust:
> Cannot npm install from a clean state, unable to resolve zone.js version:

@Maximilian-R Yeah, those dependency conflicts seem impossible to get rid of in a monorepo unfortunately. I've upgraded everything to their latest version, but various packages still have conflicting sub-dependencies.
It seems like the only solution is to use `npm i --force` when installing with NPM. Yarn already does this by default.

vsjolander (approved):
Tested this branch locally on MBP and all tests and storybooks worked.

Also removed all e2e apps that aren't really used. I ran test and storybooks after and they worked still.

Good job! 🎊

PR #808, opened 2023-03-23
feat: add component filter-chip

astrit:
This is a GitHub pull request (PR) for a component called "Chips". The PR includes the following changes:

-  The `handleChipSelection` and `handleChipRemoval` functions have been imported from the `config` module.
-  The `Template` function has been defined to generate the HTML for a chip element based on the provided props.
-  The `Meta` component has been used to set the title and parameters for the Storybook story.
-  The `Canvas` and `Story` components have been used to render the chip element with different props and variations.
-  The structure and variations of the chip element have been described in the documentation.
-  The accessibility features of the chip element have been documented, including keyboard and screen reader support.
-  Examples of how to use the chip element in HTML and how to add a ripple animation have been provided.
-  Examples of using multiple chips and supporting right-to-left languages have been included.

The PR provides a well-documented and tested component that can be easily integrated into a project.


sebopensource:
👋 Thanks for creating a pull request! 🚀 Checkout the storybook we've created for it: <ul><li><a target="_blank" href="https://sebgroup.github.io/green/pr-808/chlorophyll/">chlorophyll</a></li></ul>

libs/chlorophyll/scss/components/chips/_mixins.scss
```diff
@@ -0,0 +1,198 @@
+@mixin global {
```
splashdust, line 1: I like this pattern of grouping variables in a mixin! Maybe we should have a naming convention for variable mixins? Like `global-vars`, `prop-vars`, `icon-vars`, etc, in this example. That way, when you read the code which calls the mixin, you can easily tell that the `@include` statement is only outputting variables.

In a similar fashion, the style mixins in this example could be named `state-styles`, `variant-styles`, etc.

What do you think?

libs/chlorophyll/scss/components/chips/_index.scss
```diff
@@ -0,0 +1,34 @@
+@use '../../tokens/shame' as tokens;
+@use 'mixins';
+
+:root {
+  @include mixins.global-vars;
```
vsjolander, line 5: These variables should be moved to the following places I think to work with global import and the .use-green mixin:
[Global scope](https://github.com/sebgroup/green/blob/main/libs/chlorophyll/scss/components/themes/_index.scss#L81)
[use green scope](https://github.com/sebgroup/green/blob/main/libs/chlorophyll/scss/components/use-green-scope/_index.scss#L17)

vsjolander (changes requested):
Awesome work! 🤩 

Found some things that needs fixing:

- Need to move vars to two seperate places, see comment in code: [here for global scope](https://github.com/sebgroup/green/blob/main/libs/chlorophyll/scss/components/themes/_index.scss#L81) and here for [use green scope](https://github.com/sebgroup/green/blob/main/libs/chlorophyll/scss/components/use-green-scope/_index.scss#L17)
- And also noticed that the focus is not visible when the item is selected:
https://user-images.githubusercontent.com/11420341/227555509-d3798fb6-8eda-4f6f-baab-3bd5338a636a.mov
- I think that when focusing the removable chip the cross should be highlighted with a white border. Maybe check with Jessi & Jonathan.
- Icons should be white is light mode and blue in dark mode I think

Also feel that the hover transition is a bit slow perhaps? But we have to look at transitions as a whole so keep it as is 👍

Some minor details but over all super good 👌👌👌

vsjolander (approved):
Really good job @astrit 🎉🎉🎉

PR #821, opened 2023-03-31
fix: use fully qualified import for `swiped-events`

splashdust:
Some people where reporting issues compiling extract with Webpack due to a sub-dependency called `swiped-events` failing to import properly. This change should hopefully fix that issue.

@astrit FYI!


sebopensource:

astrit:
Perfect, I guess we can check if this does the job on the next meeting ✌️

PR #828, opened 2023-04-04
feat(slider): add support for unit field

### LINKED ISSUE #819: Slider - Support for showing a unit inside the input field

- [X] I confirm that I have checked if the component already exists in any of the Green libraries

React

### Description

For example "kr/mån" and "år". It should work and look the same as in the input components.

_No response_

### Anything else?

_No response_

### LINKED ISSUE #820: Slider: Input field is shown one it's own row on small/medium viewports
### Bug already reported ?

React, Chlorophyll

### Component name

Slider

### Description

Normally the input field is shown on the same row as the label, but on tablet sized screens and smaller the input field is shown on it's own row.

### Steps To Reproduce

Test here:
https://sebgroup.github.io/green/latest/chlorophyll/?path=/story/components-slider--slider

### Current Behaviour

image

### Expected Behaviour

You have to verify this with CX, but I believe it should look like this. 

image

Or there might be a breakpoint, but that should be much smaller than the current one.

splashdust:
This PR adds support for unit field to the input field in the slider component. It also removes the current breakpoint that cause the input field to render on a separate row, since that behaviour deviates from the [current design](https://designlibrary.sebgroup.com/components/component-slider)

It also refactors the SCSS slightly to better follow Green conventions.

Closes #820
Closes #819


vsjolander (approved):
Looks good to me! Well done :)

sjkayle (approved):

PR #852, opened 2023-04-21
feat: 🎸 Context Menu

sixskulls:
Minimal Context Menu, with ellipsis button to open.
TODO:
- Add ng-template to be able to add anything instead of button.
- Allow to open on left click?
- Allow to open not only down, but up as well and calculate top and left accordingly.
- Design can be imroved for mobile.


libs/angular/src/lib/context-menu/context-menu.component.html
```diff
@@ -0,0 +1,31 @@
...
+  [style.top]="top"
+  [style.left]="left"
+  #contextMenuPopover
+>
+  <ul role="listbox">
+    <li
+      *ngFor="let menuItem of menuItems; let i = index"
+      (click)="onItemClick(menuItem)"
+      role="option"
+      (keydown)="onMenuItemKeyDown($event, i)"
```
nerjok, line 17: this could pass menuItem instead index,  by which menu item is retrieved in ts file

libs/angular/src/lib/context-menu/context-menu.component.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }
+
+  open(event: MouseEvent): void {
+    if (this.isActive) {
+      this.close()
+      return
+    }
+
+    const button = this.getButtonElement(event)
```
nerjok, line 52: i would recommend update to ViewChild(), as it always stays within component.

libs/angular/src/lib/context-menu/context-menu.component.ts
```diff
@@ -0,0 +1,114 @@
...
+  ViewChild,
+} from '@angular/core'
+import { DropdownOption } from '@sebgroup/extract'
+
+@Component({
+  selector: 'ngg-context-menu',
+  templateUrl: './context-menu.component.html',
+})
+export class NggContextMenuComponent {
+  @Input() direction: 'ltr' | 'rtl' = 'ltr'
```
nerjok, line 18: maybe could be opened to different directions(to top/left) if doesnt fit into screen.

libs/angular/src/lib/context-menu/context-menu.component.ts
```diff
@@ -0,0 +1,114 @@
...
+  @ViewChild('contextMenuPopover', { static: false }) popover:
+    | ElementRef
+    | undefined
+  isActive = false
+  top = '0px'
+  left = '0px'
+
+  constructor(private elementRef: ElementRef) {}
+
+  @HostListener('document:click', ['$event'])
```
nerjok, line 34: `@HostListener('document:click', ['$event.target'])`

libs/angular/src/lib/context-menu/context-menu.component.ts
sixskulls: It could but would complicate the calculation and add edge cases, this can be developed with later version when someone needs this functionality

splashdust (changes requested):
Hey @sixskulls, thank you for the pull request!


Overall it looks really good, but I have two small nit-picks:

1. `role=option` may not be ideal for this component. The context menu (or fold-out as it is called in Design Library) is primarily intended as a way to condense a set of actions into a contextual menu. As such, the choices inside the list box should probably be buttons or links.
https://designlibrary.sebgroup.com/components/component-foldout \
\
`role=option` signals to the user that the item is a selectable option (like a select menu in a form), and it's the pattern we use in the Dropdown component. As such, I think it would be better if this one defaults to having button elements instead.

2. The hover style should not have the checkmark and bold text. I suspect this styling comes from the Dropdown styles, but it should only apply to selected items. The hover state should only have the grey background.

Again, sorry to nit pick! Apart from these small things it looks great.

splashdust (approved):
Thank you for the changes, looks good now! 🙌

PR #853, opened 2023-04-24
feat: 🎸 Angular in-page wizard step card

Fredrik-Oberg:
* [x]  I confirm that I have checked if the component already exists in any of the Green libraries
 
 Angular
 
 ### Description
 The In-page wizard step card as a Angular component
 
image

 _No response_
 
 ### Anything else?
Added `disableNext` option for disabling next button if validation of input values fails.


libs/angular/src/lib/in-page-wizard-step-card/in-page-wizard-step-card.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+        'in-page-wizard-step-card-title'
+      )
+      expect(title).toBeDefined()
+      expect(title.textContent).toEqual('Test Title')
+
+      const stepText = await fakeComponent.findByTestId(
+        'in-page-wizard-step-card-step-text'
+      )
+      expect(stepText).toBeDefined()
+      expect(stepText.textContent?.trim()).toEqual('Test Step Text')
```
Fredrik-Oberg, line 45: Noticed a weird bug that I suspect is connected to prettier setup. 
The prettier is formatting the HTML in a way that it adds a white-space before and after text that is directly under a div element. Did not see the same thing when using vs-code command "Save without formatting" and having the HTML in one line.

https://github.com/sebgroup/green/pull/853/files#diff-0f05445b29f895bce4dd447966508fea7a3ec664b8e9f6d84764572efe2b2826R20

Fredrik-Oberg:
@splashdust Updated with tests

splashdust:
@Fredrik-Oberg Thank you for the added tests!

I made a small change to the code, renaming the module to simply `InPageWizardModule`, since we may want to extend the functionality to include a complete generalized wizard implementation in the future. In that case I think the `InPageWizardStepCardComponent` should be one of a handful of components exported by that module.
Let me know if you disagree!

Otherwise everything looks good to me, and this PR is good to merge.

splashdust (approved):

PR #856, opened 2023-04-30
feat: 🎸 Add handle edit click event emitter

Fredrik-Oberg:
* [x]  I confirm that I have checked if the component already exists in any of the Green libraries
 
 Angular
 
 ### Description
Added missing event on edit button click. 
This is needed when content inside component needs to change state based on when edit button is clicked.
 
For example to an input that should be disabled when step card is not active. 

 _No response_


splashdust (approved):
Nice work, and thank you for the PR!

PR #859, opened 2023-05-02
fix: 🐛 Missing Output decorator to wizard step card

Fredrik-Oberg:
Add Missing Output decorator to in-page wizard step card


splashdust (approved):
💯

PR #876, opened 2023-05-10
fix: 🐛 Make wizard step card footer button show flex-start

Fredrik-Oberg:
The footer button in wizard step card is showing to the right side. 
This is because the card footer CSS is overriding the wizard step card footer CSS.

Scoped the wizard step card footer to show flex-start.

Before:
image

After:
image


splashdust (approved):
Looks good, thank you!

PR #880, opened 2023-05-10
Fix/context menu fixes

sixskulls:


nerjok:
> Looks good, thank you for the PR! 🥇

Hi, does anyone test components before accepting to green?  This only works then it in the page center, in many cases it will be hidden or broken with positioned elements. Even more introduced additional directive add excessive functionality, not even looking at tests... Please look at contributions more carefully.

splashdust:
@nerjok We have limited ability to do any in-depth rigorous testing, but we do check that design guidelines are followed, that there aren't obvious omissions in accessibility, that tests exist and pass, and so on.

If a pull request is opened, it usually means that it contains code that works for at least one use case, and if it passes some sanity checks we'll merge and deploy it as soon as possible, but that's about it. It does not have to be perfect of feature-complete. If bugs or performance issues are discovered, or if the need for additional features arise later on, a new PR can be opened with those additions.

So with that said, thank you for voicing your opinion and engaging with the development of Green! It sounds like you have some ideas for how to improve the context menu, and a PR with those improvements would be very welcome!

PR #888, opened 2023-05-12
chore: storybook cash fix for seb pc 

EldRoos:
Remove storybook cashing so it building works on SEB pc.
(Ok for others to get this change too as it is not working optimally anyways).


splashdust (approved):
Good stuff! I just moved it to the root config, since it should affect all storybooks, and added a check to options.cache, since it seem to not exist in all environments.

PR #891, opened 2023-05-15
fix(icon-button): add missing type

patrikfogelstrand:
The type attribute was missing from the IconButton, meaning if the button was used inside a `<form>`, it would [act as a submit button by default](https://stackoverflow.com/a/9824845/12416767). This meant that the expandableInfo button would trigger submission of a form, instead of just toggling the expandable info.

Adding it as a optional property, which defaults to `button` not to cause any breaking changes.

### DIFF
--- a/libs/chlorophyll/scss/components/button/_index.scss
+++ b/libs/chlorophyll/scss/components/button/_index.scss
@@ -56,6 +56,7 @@ button[type='button']:where(
         .danger,
         .close,
         .link,
+        .icon,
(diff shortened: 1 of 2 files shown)


patrikfogelstrand:
Adding the type gives the wrong hover and active states. Is the styling based on the type also? 

hover: 
image

active: 
image

desired hover:
image

desired active: none (same as hover)

splashdust (approved):
Looks good! Thanks for the PR! 💯

PR #893, opened 2023-05-16
chore: update authors

patrikfogelstrand:
SEB generic packages should use a SEB generic author.


splashdust (approved):
Yes, that makes sense. Good catch!

Issue #877, opened 2023-05-10
List: Switch label and value boldness


EldRoos:
### Bug already reported ?

- [X] I confirm that I have checked if the bug already has been reported

### For which framework/library you are reporting the bug

Chlorophyll

### Component name

List

### Description

The list group component in Green seems to have the label in medium and the value in regular. In the Figma component it's the opposite.
/Theresa Hedlund

### Steps To Reproduce

Compare:
https://sebgroup.github.io/green/latest/chlorophyll/?path=/docs/components-list--page
and:
https://designlibrary.sebank.se/components/component-list

### Current Behaviour

Label: medium
Value: regular

### Expected Behaviour

Label: regular
Value: medium

PR #896, opened 2023-05-19
fix: lists font wight and size

EldRoos:
#877
In vale lists and list groups: Removed inverted class, and inverted all font weights on label and value - as the Figma sketch. Also changed header to normal weight and size 14px (from bolder and smaller).


EldRoos:
Should a debug be added saying inverted class is depricated?

splashdust (approved):
Looks good!

Just for reference, if anyone wonders why the rem values are now hard coded: We decided that it's OK to hard-code them here for now, because we are re-engineering the way we work with design tokens for the 2023 design.

PR #897, opened 2023-05-22
fix: 🐛 Context menu on scroll markForCheck and css fixes

sixskulls:
Added markForCheck due to angular change detection not caching changes on scroll and improved css

BREAKING CHANGE: 🧨 -

✅ Closes: -


splashdust (changes requested):
Use utility class instead of `.small-padding-adjusted`. See comment on the scss code.

splashdust (approved):
👍

Issue #373, opened 2022-01-28
Tab: New component - Tab navigation


ulde01:
### Design decision
2022-01-27
![image](https://user-images.githubusercontent.com/45568194/151574315-ef709617-2ff7-4bc7-8d1a-979527c3e4d5.png)

### What is this
Tab navigation lets users to explore and switch between different content in our channels. The tabs are tied to the site navigation.

### Details
https://designlibrary.sebgroup.com/components/component-tabnavigation

### Component identifier
component-tabnavigation

### Follow up - development

Let me know if you have questions. And ss always, please comment and close if this issue is not relevant to you.
/Ulrika

Todos:
- [ ] Markup
- [ ] Tokens
- [ ] Story

vsjolander:
These type of component should be "dumb" so the application handles things like routing.

vsjolander:
This is more of a styling variant of tabs: https://sebgroup.github.io/green/latest/chlorophyll/?path=/docs/components-tabs--tabs

PR #903, opened 2023-05-30
feat: 🎸 components can be added as tab content

comeab:
Currently tabs only do not allow any content other than text. Second issue, the tab content and tab title are always equal. Now the tabs component allows adding tab children  components with any type of content

BREAKING CHANGE: 🧨 Tabs component now takes Tab child components as children.

✅ Closes:  #373

### DIFF
--- a/libs/react/src/lib/tabs/tabs.spec.tsx
+++ b/libs/react/src/lib/tabs/tabs.spec.tsx
@@ -1,7 +1,60 @@
 import { render, screen, fireEvent } from '@testing-library/react'
-import Tabs, { Tab, TabProps } from './tabs'
+import Tabs, { IList, Tab } from './tabs'
 
-const list: TabProps[] = [
+
+const list: IList[] = [
+  { text: 'Page 1', href: '#' },
+  { text: 'Page 2', href: '#' },
+  { text: 'Page 3', href: '#' },
+  { text: 'Page 4', href: '#' },
+  { text: 'Page 5' },
+  { text: 'Page 6', disabled: true },
+]
+
+describe('Tabs only allow text as content', () => {
+  it('should render all anchor elements', () => {
+    render(<Tabs list={list}></Tabs>)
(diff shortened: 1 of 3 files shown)


splashdust:
Hey @comeab! Thanks for the PR!

When refactoring components we need to be mindful of breaking changes, and it's good that you have clearly noted the breaking change in this PR. However, we still want to avoid making breaking changes unless strictly necessary, so the preferred compromise here is to *deprecate* the old feature, but not remove it right away. That gives consumers of the framework a heads-up that a breaking change is coming in the future, and gives them a chance to refactor their code in advance.

So in this case, I think we need to keep the old `list` property, while still adding support for the `Tab` children. If no `Tab` children are present, the component should fall-back to the deprecated render method.

Then, in the `TabsProps` interface, add a JSDoc comment to the `list` field, something like this:
```typescript
/**
 * @deprecated use `<Tab>` child components instead
 */
 ```
That way, any current consumers of the Tabs component will get a compile-time warning that this part of the API has been deprecated.
 
But with that said, I think this is a great improvement. Using `Tab` children is a much more natural interface for a React component!

comeab:
Hi @splashdust 
Fair comment. In fact I thought about this but decided to give it a shot as is. The reason being that the current implementation was somewhat unusable for the purpose of a tab. You cannot really supply any content to the tabs only the titles which are then set as content by default from the IList. So  I concluded "assumed" that it most likely was not being used yet but expected that this could be an issue.

I have now restored the previous implementation and the tests. And will certainly follow this approach for next PRs

Thanks

splashdust:
@comeab Thank you for the update, awesome work! 💯 

I agree that in this case it's fairly likely that no one was using the previous implementation, but we can't really know for sure, and as Green keeps maturing and getting more widely used, it becomes even more important to follow these kinds of principles, even if it can seem silly at times!

Feel free to mark the PR as 'Ready for review' when you are otherwise happy with it!

Issue #904, opened 2023-06-01
Tab: only render tab content of selected tab


comeab:
### Description

Currently all tab contents are rendered all at once. In cases where tab content has components that do network calls for instance to fetch data or any other IO work or heavy calculations then even if the user does not select that tab during user journey, the content will have been loaded anyway. This would reduce calls to the server and unnecessary rendering.

### Who will benefit from this feature?

Currently developing a Microfront that loads data to multiple tabs from the server. The data should only be loaded when the user selects the specific data.

PR #905, opened 2023-06-01
perf: ⚡️ only renders tab content when tab is selected

comeab:
When tabs have components making network calls or any heavy processing, it might be ideal to delay this processing until the user has selected that tab. In most cases with tabs, users will not be clicking aroound eveyry tab all the time and so loading such content creates unnecessary overhead

BREAKING CHANGE: 🧨 None

✅ Closes: #904


splashdust (approved):
Great stuff, thank you for the PR!

Issue #923, opened 2023-06-28
Slider: Textbox has no associated label


daneriseb:
Slider (React) - textbox (number input) has no associated label

### Bug already reported ?

- [X] I confirm that I have checked if the bug already has been reported

### For which framework/library you are reporting the bug

React

### Component name

Slider

### Description

When testing Accessibility with Lighthouse I get these kind of errors:
![image](https://github.com/sebgroup/green/assets/113447576/4e848ba5-c4b3-4df9-989e-7bc9a1cccb4f)

There is a label, but it's only associated to the range input element.

![image](https://github.com/sebgroup/green/assets/113447576/1c94dafc-12dc-40a3-8d75-0273722402bf)

I see the same problem in the Green storybook.

Issue #920, opened 2023-06-22
Slider: No mobile version


EldRoos:
### Bug already reported ?

- [X] I confirm that I have checked if the bug already has been reported

### Component name

Slider

### Description

No mobile version of the slider (with full width input), see Figma. (after figma changes the input should be normal sized on mobile and small on desktop).

### Current Behaviour

Just makes everything more narrow.

### Expected Behaviour

Should fill full width in separate rows.

EldRoos:
Somewhat connected to this issue: #866

Issue #902, opened 2023-05-29
Slider: Input field is too high


daneriseb:
### Bug already reported ?

- [X] I confirm that I have checked if the bug already has been reported

### For which framework/library you are reporting the bug

React

### Component name

Slider

### Description

According to "SEB Design System" in Figma the slider input field should have a height of 32px. Now it is 48px.

EldRoos:
I confirm Figma and Green have different sizes (though it is 44px not 48px.

Will check if at least the mobile version should have normal bigger sized input field.

PR #929, opened 2023-07-04
fix(slider): mobile version, small input on desktop, added aria-lable

EldRoos:
#902 #920 #923

### DIFF
--- a/libs/angular/src/lib/slider/slider.component.html
+++ b/libs/angular/src/lib/slider/slider.component.html
@@ -42,6 +42,7 @@
     [attr.name]="name"
     [attr.id]="name + '-textbox'"
     [attr.placeholder]="placeholder"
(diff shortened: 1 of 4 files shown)


EldRoos:
is aria-label="Slider value" good enough, or should the label to be able to be set? (or should a hidden label be used instead?)

splashdust:
@EldRoos Good question. One problem I can see with `aria-label` here is that it will be hard coded to an english label, which will not make sense in other languages.

Could we put an `id` on the existing `<label>` instead, and then use `aria-labelledby` on the input?

Simplified example:
```html
<label for="slider-id" id="label-id">Label text</label>
<input type="number" aria-labelledby="label-id">
<inpup type="slider" id="slider-id">
```

splashdust (approved):
Looks good! 👍

PR #930, opened 2023-07-05
feat: 🎸 new input to set fixed placment of dropdown options

sjkayle:
### Issue
There is a case where we need the dropdown options to be fixed at the top and not flip to the bottom (flipping to the bottom _cuts_ the dropdown options below):

Media2

### Fix
Add a new attribute to accept whether the dropdown options placement should be fixed either at the top or the bottom.
i.e., passing "top-start" would fix the options at the top and disable the flipping:

image

### DIFF
--- a/libs/extract/src/lib/dropdown/dropdown.ts
+++ b/libs/extract/src/lib/dropdown/dropdown.ts
@@ -227,8 +227,6 @@ const pop = (
       })
     }
 
-    console.log(_popperModifiers)
-
     handler.popper = createPopper(handler.toggler, handler.listbox, {
       placement: fixedPlacement ?? 'bottom-start',
       modifiers: _popperModifiers,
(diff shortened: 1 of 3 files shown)


libs/extract/src/lib/dropdown/dropdown.ts
```diff
@@ -197,24 +200,38 @@ const update = async (
...
+      _popperModifiers.push({
+        name: 'flip',
+        options: {
+          allowedAutoPlacements: [],
         },
-      ],
+      })
+    }
+
+    console.log(_popperModifiers)
```
eweseong, line 230: this PR is merging into main, is the console logging here intended?

sjkayle: oops, removed 😅

splashdust (approved):
Looks good to me, nice work! 👍

PR #954, opened 2023-08-08
fix: 🐛 add onchange support to form items component

eweseong:
provide an alternative method for form context consumer to handle input change event in addition to the existing of `onFormSubmit` handler that is only trigger upon form submission

### DIFF
--- a/libs/react/src/lib/form/formItems.tsx
+++ b/libs/react/src/lib/form/formItems.tsx
@@ -81,7 +81,7 @@ export const FormItems: React.FC<FormItemsProps> = ({
     name,
     onChange: (event: React.ChangeEvent<HTMLInputElement>) => {
       if (typeof _onFormItemsChanged === 'function') {
-        _onFormItemsChanged?.(event)
+        _onFormItemsChanged(event)
       }
 
       onChange(event)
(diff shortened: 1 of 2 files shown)


sjkayle (approved):

mario-subo (approved):

terrance456 (approved):

libs/react/src/lib/form/formItems.tsx
```diff
@@ -48,21 +53,39 @@ export const FormItems: React.FC<FormItemsProps> = ({ children, validate, name }
...
 
   /* eslint-disable-next-line */
-  return React.cloneElement(children as any, {
+  return React.cloneElement(children as React.ReactElement, {
     validator: errors?.[name] && validate,
     name,
-    onChange,
+    onChange: (event: React.ChangeEvent<HTMLInputElement>) => {
+      if (typeof _onFormItemsChanged === 'function') {
+        _onFormItemsChanged?.(event)
```
splashdust, line 84: No need for optional chaining here, since the `if` already ensures this is a function value :)

splashdust (changes requested):
Just one small nit-pick (see comment), otherwise looks good!

PR #962, opened 2023-08-16
Green Core

splashdust:
# Green Core

***Note to reviewers:** I don't expect an actual review of the changes in this PR. There is a lot! If you're tagged as a reviewer, that basically just means that you've been involved in Green in one way or another, as a contributor or user, and I wanted to get your attention :)*

*If you have questions, concerns or suggestions about the move to Web Components in Green, now is a good time to raise them!*

## About Green Core

Green Core is a carefully crafted set of native Web Components, laying the foundation of the Green Design System.

The aim of the Core package is to provide a solid, accessible and maintainable foundation for the Design System. Using Web Components allows us to encapsulate styling, markup and logic in a framework agnostic way. It also allows us to solve accessibility challenges without tripping over the different template engines and DOM management strategies employed by various frameworks.

Green Core breaks components down into "Primitives". Primitives are component building blocks that can be composed together to build complete components. For example, the Dropdown composes `gds-popover` and `gds-listbox` to construct a complete Dropdown.
The primitives encapsulate appropriate semantics and accessibility behaviour, and are used by multiple components. This approach promotes reuse and consistency throughout the Design System.

The shape of the overall architecture will look something like this:

```mermaid
flowchart TD
    DT[Design Tokens]---S
    subgraph "Green Core"
        S[Styles]---AP
        M[Markup]---AP
        AP[Atomic Primitives]---|Composition|CP[Compound Primitives]---|Public API Boundry|WC["Web Components"]
    end
    WC---A[Angular]
    WC---R[React]
```

There will be three main options for consuming components: As native Web Components or through React or through Angular. The React components will essentially just be thin wrappers over the native ones, and Angular can use the WC's directly, as it has excellent support for that. Some components will have wrapper components in Angular as well though, in order to avoid too many breaking changes.

Eventually, Core will replace the current Extract package.

### Challenges

Web Components does introduce some challenges. One challenge is that there is only a single `CustomElementsRegistry` in the browser, and Green Core needs to work seamlessly in a Micro-Frontend environment, where multiple instances of the framework can operate in parallell. Normally, this would lead to conflicts, when two instances try to register the same element name. Green Core solves this by introducing a transparent name-scoping mechanism. By default, element names in Green Core will get a version determined suffix applied to their element name. This happens automatically by the templating engine, so developers should not have to think about it most of the time.

More info about scoping can be found here: https://github.com/sebgroup/green/blob/4155876f787463c904c753f07bb8a78563a0f90e/libs/core/src/utils/helpers/custom-element-scoping.ts

## About this PR

This PR adds the initial implementation of Green Core, and contains the basic set up of the new package as well as one mostly finished component, the dropdown.

~~The Web Component based Dropdown will replace the current Extract based dropdown in Angular and React when this PR is merged. It should be mostly backwards compatible, but since this is a complete rewrite, some breaking changes can be expected and some aspects of the existing API will be deprecated.~~

A separate PR will be opened to replace the current Angular and React implementations, so initially, the release of Green Core will not affect any of the existing packages.

This is a partial merge of the ongoing work in the `vision` branch. If you are curious about this work you can take a look at the `vision` branch. For now, the bulk of the work will continue in that branch, but as more of the work matures, we will start doing feature PRs directly against `main`.

This PR is just the first step towards migrating Green to the new Web Component based architecture.

## Todo
Some work remains before this PR can be merged:
- [x] Fix validation states of the Dropdown in Angular
- [x] Finish scoping implementation by generating an actual version-based suffix (currently it is just a hard-coded constant)
- [x] Fix failing tests and linting across all libs


splashdust:
Storybooks now exist for this PR, though the automated comment action failed due to missing config.

Here they are:

Core: https://sebgroup.github.io/green/pr-962/core
React: https://sebgroup.github.io/green/pr-962/react
Angular: https://sebgroup.github.io/green/pr-962/angular

patrikfogelstrand:
@splashdust 

On parental leave until Jan-24, but took a look at this now, overall it looks really promising, great job. 

Really cool solution with custom element scoping, looking forward to dig into it once I'm back.

astrit (approved):

PR #970, opened 2023-08-21
fix: 🐛 fix context menu position in relative parent

nikolajus-seb:
When context menu was put in a parent HTML element, it was displayed in a wrong position due to top and left coordinates being calculated from the whole window, not from parent coordinate system. This change introduces an overlay in order to mitigate the issue in the simplest possible manner. The solution was inspired by material angular library tooltip component implementation.

BREAKING CHANGE: 🧨 -

✅ Closes: -


sixskulls (approved):

splashdust (approved):
Looks good! 👌

PR #973, opened 2023-08-22
fix: 🐛 wrong calculation of y position for context menu

nikolajus-seb:
Due to the previous fix for x and y positions (adding an overlay) it is no longer needed to add scroll offset to the context menu y position.

BREAKING CHANGE: 🧨 -

✅ Closes: -


sixskulls (approved):

splashdust (approved):
👍

PR #974, opened 2023-08-24
fix(typography): switch back to unitless `line-height` for body

splashdust:
The current fixed rem based unit cause issues in situations where font-size is enlarged.


vsjolander (approved):
Looks like a mistake from my end, can't remember any specific reason for this.

PR #975, opened 2023-08-25
fix(react): fixed radio group prop type expandableInfo

vsjolander:
Extracting properties to its own interface to ensure consistency of this functionality. RadioGroup and Inputs had different interfaces for expandable info props. They should now extend the IExpandableInformation.


astrit (approved):
Looks good to me 🤟

splashdust (approved):
Good stuff! 👍

PR #976, opened 2023-08-28
fix: 🐛 wrong position after zoom-in due to cached position

nikolajus-seb:
When zooming in / out the component does not reset it's position to (0; 0), hence when no width (only max-width) is specified the initial boundary condition on how to display it when it is close to the border of the container does not fully suffice and produce incorrect dynamic width (open and close differs in the context menu width too)

BREAKING CHANGE: 🧨 -

✅ Closes: -


sixskulls (approved):

splashdust (approved):
Looks good! 👍

PR #987, opened 2023-09-05
Fix/in wizard

sjkayle:
Changes include -

- use correct font-weight keyword for strong tag/class (semibold is invalid)
- update documentation for in-page wizard summary step to make subtitle bold as per design library guidelines

### Preview

image

### DIFF
--- a/libs/angular/CHANGELOG.md
+++ b/libs/angular/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 This file was generated using [@jscutlery/semver](https://github.com/jscutlery/semver).
 
+# [1.9.0](https://github.com/sebgroup/green/compare/@sebgroup/green-angular@1.8.5...@sebgroup/green-angular@1.9.0) (2023-09-01)
+
+
+### Features
+
+* **angular:** modal customizable header and footer ([#982](https://github.com/sebgroup/green/issues/982)) ([cdc1668](https://github.com/sebgroup/green/commit/cdc1668013316459c36d13f55a6cc981e16f377b))
+
+
+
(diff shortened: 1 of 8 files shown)


mario-subo (approved):

vsjolander (commented):
Thanks for the PR! 

I'm not sure about this though. We did change the style of headers which changes this design since we use a `h4` in the example. 

Adding a `.strong` class is more of a typography feature that might be needed but I think we neeed to think more about these utility classes.

I think changing the content of this wizard example to a [value list component](https://sebgroup.github.io/green/latest/chlorophyll/?path=/docs/components-list--page#value-list) would be better. We did have a inverted variant of that component that is exactly what we were looking for but [it was removed](https://github.com/sebgroup/green/commit/d36caac913986b3f66084707e3697ef8070cacc6). It [exsists in Figma](https://www.figma.com/file/dX0HJ2MOaQyAOTziCD5fQ7/SEB-Design-System?type=design&node-id=14694-51365&mode=design) but it lives in a weird place (Inputs?).

We should probably move the Figma component to lists and call it value list with the variant of inverted or smth and reimplement the css and change the content of this wizard example or skip this all together... not sure..

What do you guys think @splashdust @astrit?

sjkayle:
@vsjolander using a value list seems like a good idea

splashdust (approved):
👍

splashdust:
@vsjolander @sjkayle Didn't notice the comments until after I had left review, but yes, I agree on changing it to a value list!

sjkayle:
@vsjolander @splashdust I've pushed a commit and used value lists in the documentation as suggested. Kindly review again.

image

splashdust (approved):
Great!

Also, I didn't notice the addition of `.strong` at first, but I removed it now. There is already a utility class called `.fw-bold` that can be used to change font-weight. So if you need bold text without *strong* semantics, that's the one you should use.

PR #990, opened 2023-09-06
feat(react): information label can now be react node

vsjolander:
Adding the possibility to pass a react node to the information label. added some overriding temporary styling since it clashed with old generated styling based on key words.

Also extracted the label and label information types to it's own interface.

This was requested by the loan team.

### DIFF
--- a/libs/extract/src/lib/form.ts
+++ b/libs/extract/src/lib/form.ts
@@ -6,3 +6,8 @@ export interface IExpandableInformation {
   expandableInfo?: ReactNode
   expandableInfoButtonLabel?: string
 }
+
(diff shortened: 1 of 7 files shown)


libs/extract/src/lib/form.ts
```diff
@@ -6,3 +6,8 @@ export interface IExpandableInformation {
   expandableInfo?: ReactNode
   expandableInfoButtonLabel?: string
 }
+
+export interface ILableAndLableInformation {
```
splashdust, line 10: There's a spelling error in the interface name I believe, "lable" should probably be "label".

PR #991, opened 2023-09-07
fix: 🐛 pagination shouldn't exceed totalPages on length change

nikolajus-seb:
When on the last page and changing length (for example, deleting items) the pagination component should react accordingly and, if needed, move user to a correct page.

### DIFF
--- a/libs/angular/src/lib/pagination/pagination.component.ts
+++ b/libs/angular/src/lib/pagination/pagination.component.ts
@@ -71,6 +71,10 @@ export class NggPaginationComponent implements OnChanges {
     ) {
       this._pageIndicies = this.getDisplayedPageIndicies()
     }
+
+    if (changes.length && this.pageIndex >= this.totalPages) {
+      this.goto(this.totalPages - 1)
+    }
   }
 
   hasPrevious(): boolean {


vsjolander:
Hey and thanks for the PR! 👊 

I would approve and merge this directly if it weren't for commitizen to mark this as a breaking change which till bump the version to 2.0.0 at this point.

Not sure if you can remove "BREAKING CHANGE: 🧨 -" manually from the commit message or if you have to make a new PR with a commit that doesn't contain this.

Basically you just leave the field empty when going through the commitizen commit flow. Adding a "-" will think it contains a breaking change. Same for "Closes:".

nikolajus-seb:
Oh, sorry, didn't know about that! Removed the lines altogether right now. They have been added by git-cz (the tool that you recommend somewhere in the README), hence thought that putting '-' is appropriate when nothing breaks.

libs/angular/src/lib/pagination/pagination.component.ts
```diff
@@ -71,6 +71,10 @@ export class NggPaginationComponent implements OnChanges {
     ) {
       this._pageIndicies = this.getDisplayedPageIndicies()
     }
+
+    if (changes.length && this.pageIndex >= this.totalPages) {
+      this.goto(this.totalPages - 1)
```
nerjok, line 76: it looks like someone is hiding real problems, pagination component receives `pageIndex` and pages count `length`,  in my opinion this component shouldn't be responsible for wrong data. After this  - page change event is thrown meaning, that any its listeners would for example trigger endpoint call, that could lead to unexpected behaviour.

vsjolander: Valid point. If the length changes you should also have calculated and change pageIndex.

I'm also thinking about what causes the length to change? Filtering? Choosing a new data source? In these cases wouldn't you want to end up on the first page? Seems like a UX problem more than a component one.

nikolajus-seb: The cause is deletion, but definitely a good point to handle it on the consumer side. Closing the PR. Thanks for the review!

PR #992, opened 2023-09-07
fix(react): adding name attribute to select

vsjolander:
adding a prop so the name attribute can be set on the select component


splashdust (approved):
👍

Issue #922, opened 2023-06-28
In-page Wizard: No edit button on mobile view


sjkayle:
### Bug already reported ?

- [X] I confirm that I have checked if the bug already has been reported

### For which framework/library you are reporting the bug

Chlorophyll

### Component name

In-page Wizard

### Description

On mobile view of the in-page wizard card, the edit button is missing:
![image](https://github.com/sebgroup/green/assets/82629695/f35c3b72-89b4-4d21-a650-cce77c87cd25)

Desktop view has the edit button on the top right corner:
![image](https://github.com/sebgroup/green/assets/82629695/551c35a7-a501-423e-9451-43605484d7cb)

### Current Behaviour

No edit button on mobile view

### Expected Behaviour

Should have the edit button to re-open the step card

PR #1014, opened 2023-09-16
fix: 🐛 edit button on mobile view

sjkayle:
### Issue

Closes: [#922](https://github.com/sebgroup/green/issues/922)

#### Completed, editable step

image

#### Completed, non-editable step

image

### DIFF
--- a/libs/chlorophyll/scss/components/wizard/in-page/_mixins.scss
+++ b/libs/chlorophyll/scss/components/wizard/in-page/_mixins.scss
@@ -2,6 +2,7 @@
 @use '../../../tokens';
 
 @mixin add-in-page-wizard {
+  $prefix: &;
   border: 0;
   margin-bottom: 0.5rem;
   padding: 0;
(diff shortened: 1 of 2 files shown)


mario-subo (approved):

libs/chlorophyll/scss/components/wizard/in-page/_mixins.scss
```diff
@@ -123,4 +115,30 @@
...
     background-color: tokens.get(tokens.$green, 1);
   }
+
+  &.completed {
+    &.editable {
+      #{$prefix}__header__edit {
+        grid-area: edit;
+        display: none;
+        justify-content: end;
+        @media (min-width: 576px) {
```
splashdust, line 125: Hey, thank you for the PR!

Here I think we should use one of the predefined media query mixins, something like:
```scss
@include common.media-breakpoint-up(sm) {
  display: flex;
}
```
Same thing applies a few lines down in this file.

Otherwise looks great!

splashdust (changes requested):

sjkayle:
@splashdust refactored to use a pre-defined mixin

splashdust (approved):
👌

sjkayle:
@splashdust there seems to be a bug here when we were testing out the component. will analyze it a bit more and then raise another pr soon

Issue #863, opened 2023-05-05
Dropdown: Works poorly in NVDA


splashdust:
### Bug already reported ?

- [X] I confirm that I have checked if the bug already has been reported

### Component name

Dropdown

### Description

NVDA fails to announce the names of options in the dropdown. It might be because there are two nested `[role=listbox]` elements in there, but needs further investigation.

splashdust:
Completed in #1016

Issue #862, opened 2023-05-05
Add backdrop for mobile dropdown popover


splashdust:
### Description

The mobile dropdown is currently missing the backdrop behind the popover:

![MicrosoftTeams-image (2)](https://user-images.githubusercontent.com/129963380/230096021-50c139ba-59fa-4a25-a29e-467badc30259.png)

splashdust:
Completed in #1016

Issue #779, opened 2023-03-14
Dropdown: Rebuild


vsjolander:
Dropdown component has issues with accessibility, design and performance. It is also built as a standalone "component application" (it's huge) in extract which makes it complicated for new developers/contributors to understand the component.

It should therefor be rebuild in both angular and React to make the component perform better and make it easier to maintain.

When rebuilding the component we should build a general component that handles popovers/modal/floating item placement connected to another UI element like a "List box" to a dropdown or a "Action menu" to a "More button". This component could be a Webcomponent wrapper and is used in both angular and React. But this would have to be investigated a bit more.

Perhaps the Dropdown listbox or other floating menus should be the dialog html element to more accessible?

Styling for List box items should be created in Chlorophyll and shared by using the same class name.

vsjolander:
Related: #777

splashdust:
Completed in #1016

Issue #700, opened 2023-01-12
Dropdown: Doesn't set the for property for the label


daneriseb:
According to the Webaim accessibility standards (https://webaim.org/standards/wcag/checklist#sc3.3.2) all form labels must be associated with a form control.

splashdust:
Completed in #1016

PR #1016, opened 2023-09-18
Start using Core Dropdown in Angular and React

### LINKED ISSUE #638: Dropdown: Not compatible with screen reader
When opening a dropdown it's not possible to get the list-box items to be read by screen reader.

splashdust:
This PR swaps out the internals of the Dropdown component in React and Angular with the new Web Component based one from Green Core.

A number of things needs to be done, tested and verified before this PR can be merged, check the associated [milestone](https://github.com/sebgroup/green/milestone/7) for more details.

Closes #638, #667, #700, #779, #863, #696, #600, #862


splashdust:
Storybook comment action failed, but here are the related Storybooks:

Angular: https://sebgroup.github.io/green/pr-1016/angular/?path=/docs/components-dropdown--custom-option#dropdown
React: https://sebgroup.github.io/green/pr-1016/react/?path=/docs/components-dropdown--dropdown#dropdown

splashdust:
## Green Angular 2.0.0
The two main changes in this version is that the base version of Angular have been updated to 15, and the Dropdown component have been migrated over to the new Web Component powered version from `@sebgroup/green-core`.

### Breaking changes
* Angular updated from 14 -> 15. This means that if your project is still on Angular 14, you will most likely need to upgrade to 15 in order to use this version of Green Angular.
* Any Dropdown related imports that was previously imported from `@sebgroup/extract` will now need to be imported from `@sebgroup/green-angular` instead.
* The `loop` attribute of `ngg-dropdown` no longer has any effect and should no longer be used. This was used configure wether keyboard navigation in the listbox should wrap around when you reach the last option. The motivation for removing the option is that the behaviour should always be the same and not configurable.

## Green React 2.0.0
The main change in this version is that the Dropdown component have been migrated over to the new Web Component powered version from `@sebgroup/green-core`.

### Breaking changes
* Any Dropdown related imports that was previously imported from `@sebgroup/extract` will now need to be imported from `@sebgroup/green-react` instead.
* The `loop` property of `Dropdown` no longer has any effect and should no longer be used. This was used configure wether keyboard navigation in the listbox should wrap around when you reach the last option. The motivation for removing the option is that the behaviour should always be the same and not configurable.

## Extract 2.0.0
The main change in this version is that Dropdown has now been dropped from Extract. The framework independent implementation of Dropdown is now a Web Component in `@sebgroup/green-core` instead.

### Breaking changes
* Dropdown has been dropped from Extract

vsjolander (approved):
Extensive work!! Very well done 😊  Looks good to me but I also need to get into this lib more and also into webcomponents in general.

I managed to the styles in place - we didn't have designs for listbox or dialog so will have to check this with Jessi.

splashdust:
> I managed to the styles in place - we didn't have designs for listbox or dialog so will have to check this with Jessi.

I believe there are designs for this in Figma, but it's called "Dropdown list" there. The use of `dialog` here is just a semantic choice, and shouldn't really affect styling.

PR #1042, opened 2023-10-03
chore: updating to angular 15 support

vsjolander:
Still problems with eslint and typescript that needs to be solved.


splashdust:
I've merged these changes into #1016, so closing this one. Wanted to test the changes in #1016 with Angular 15, and this way we won't have any conflicts.

PR #1043, opened 2023-10-04
feat(charts): custom colours

vsjolander:
You can now pass a [color object](https://naver.github.io/billboard.js/release/latest/doc/Options.html#.color) in the styles object in settings for the chart.

This makes it possible to override the default theme where needed.


libs/charts/src/lib/billboard.ts
```diff
@@ -248,6 +254,7 @@ export const createInfo = (
 
 export const create = ({ settings, chartElement }: ChartArgs): Chart => {
   const options = createOptions({ settings, chartElement })
+  console.log(options)
```
splashdust, line 257: Looks good, just a stray console.log here to delete before merging :)

splashdust (approved):

PR #1045, opened 2023-10-09
fix(chlorophyll): only import tokens in use-green-scope

Maximilian-R:
radio/checkbox/filter styles are added outside of the .use-green scope. 

@use '...tokens/component' causes its styles to be imported. Only the tokens should be imported.

### DIFF
--- a/libs/chlorophyll/scss/components/use-green-scope/_index.scss
+++ b/libs/chlorophyll/scss/components/use-green-scope/_index.scss
@@ -5,9 +5,7 @@ $font-path: '../fonts' !default;
 );
 
 @use '../../tokens/colors';
-@use '../../components/form/radio';
-@use '../../components/form/checkbox';
-@use '../../components/filter-chip';
+@use '../../tokens/components';
 @use '../../tokens/shape';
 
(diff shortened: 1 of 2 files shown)


splashdust (approved):
Great stuff! 👍 

This is technically a breaking change, since some consumers may not realise that their radio/checkbox/filter-chip styles are coming from the `use-green` import. So let's wait a bit before merging this. There's a few more necessary breaking changes coming up soon, so might as well release all of it at the same time!

splashdust:
For the changelog:

## Chlorophyll 2.0.0
There is one minor breaking change in this version, related to `use-green-scope`. When you import `use-green-scope` according to [the docs](https://sebgroup.github.io/green/latest/chlorophyll/?path=/docs/use-cases-use-green-for-mfe-in-shell-application--page), you get all the required CSS variables declared in the `.use-green` scope. However, in the previous version you also got the styles for certain components included here as well, which is incorrect.

### Breaking changes
* Remove styles for `radio`, `checkbox` and `filter-chilp`from `use-green-scope`. These styles should instead be imported separately.

PR #1049, opened 2023-10-10
feat(slider): change to a text input with `inputmode='numeric'`

### LINKED ISSUE #1047: Slider Input - Wrong keyboard on input field in IOS devices
### Bug already reported?

Angular

### Component name

slider

### Description

image

Since the input type is number, the keypad pattern should be numeric.

### Steps To Reproduce

_No response_

### Current Behaviour

image

### Expected Behaviour

The keypad pattern should be numeric.

splashdust:
The default keyboard for `<input type='number'>` on mobile devices is not always fitting for the Slider component. By default, iOS shows a layout with a row of numbers on top and then a collection of special characters below, and in this case the desired layout is the pure numerical keyboard.

This PR also adds forwarding of the [`enterkeyhint`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/enterkeyhint) attribute to the input field.

Closes #1047

### DIFF
--- a/libs/angular/src/lib/slider/slider.component.html
+++ b/libs/angular/src/lib/slider/slider.component.html
@@ -36,7 +36,9 @@
 
 <ng-template #inputField>
   <input
-    type="number"
+    type="text"
+    inputmode="numeric"
+    pattern="[0-9]*"
     [(ngModel)]="value"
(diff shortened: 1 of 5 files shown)


vsjolander (approved):
Nice features, nicely implemented!

Issue #988, opened 2023-09-05
AlertRibbon: wrong text color for success alert


nikolajus-seb:
### Bug already reported?

- [X] I confirm that I have checked if the bug already has been reported

### For which framework/library you are reporting the bug

Chlorophyll

### Component name

AlertRibbon

### Description

Alert ribbon success state currently has black text and icon color, while it should be white as per design library.

### Steps To Reproduce

1. Go to chlorophyll storybook: https://sebgroup.github.io/green/latest/chlorophyll/?path=/docs/components-alert-ribbon--alert-ribbon#element-structure

### Current Behaviour

Black color for success alert ribbon text and icon.

### Expected Behaviour

White color for success alert ribbon text and icon.

conandx:
It's exactly the same colour-issue with the component **Badge**

Issue #1054, opened 2023-10-13
Issue with fonts when importing style


patriklarssson:
### Bug already reported?

- [X] I confirm that I have checked if the bug already has been reported

### For which framework/library you are reporting the bug

React, Chlorophyll

### Description

When importing a style in a tsx file like this
`import '@sebgroup/chlorophyll/scss/index.scss'`

The fonts cant be found
```
GET http://localhost:4200/fonts/SEBSansSerif-Regular.woff2 net::ERR_ABORTED 404 (Not Found)
:4200/fonts/SEBSansSerif-Regular.woff:1     GET http://localhost:4200/fonts/SEBSansSerif-Regular.woff 404 (Not Found)
:4200/fonts/SEBSansSerif-Bold.woff2:1     GET http://localhost:4200/fonts/SEBSansSerif-Bold.woff2 net::ERR_ABORTED 404 (Not Found)
:4200/fonts/SEBSansSerif-Medium.woff2:1     GET http://localhost:4200/fonts/SEBSansSerif-Medium.woff2 net::ERR_ABORTED 404 (Not Found)
:4200/fonts/SEBSansSerif-Bold.woff:1     GET http://localhost:4200/fonts/SEBSansSerif-Bold.woff net::ERR_ABORTED 404 (Not Found)
:4200/fonts/SEBSansSerif-Medium.woff:1     GET http://localhost:4200/fonts/SEBSansSerif-Medium.woff net::ERR_ABORTED 404 (Not Found)
```

### Steps To Reproduce

Create a react app
in main.tsx import green styling

```
import '@sebgroup/chlorophyll/scss/index.scss'
import { StrictMode } from 'react'
import * as ReactDOM from 'react-dom/client'
import App from './app/app'


const root = ReactDOM.createRoot(document.getElementById('root') as HTMLElement)
root.render(
  <StrictMode>
      <App />
  </StrictMode>
)
```

### Current Behaviour

Wrong import

### Expected Behaviour

Correct import

vsjolander:
Hello and thanks for creating an issue!

You need to specify where the fonts are on your local machine. The default path is: `'../fonts'` and that's where its looking. So in your case you have two options: 

- Copy font files to `/fonts/SEBSansSerif-Regular.woff2`in the folder you are serving your app from. 
- Set up a `.scss` file where you import Chlorophyll with an altered `$font-path`. Change this to where you have the font files. See example below:

```scss
@use '@sebgroup/chlorophyll/scss' with (
[...]

PR #1055, opened 2023-10-13
fix(chlorophyll): text colour on success and font path

vsjolander:
Closes: #988 #1054

### DIFF
--- a/libs/chlorophyll/scss/components/utility/_colors.scss
+++ b/libs/chlorophyll/scss/components/utility/_colors.scss
@@ -2,6 +2,7 @@
 @use 'sass:map';
 
 $_background: var(--sg-bg-level-0);
+
 $colors: (
   'light': (
     color: tokens.get(tokens.$grey, 200),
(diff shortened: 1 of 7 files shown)


libs/chlorophyll/scss/components/utility/_colors.scss
```diff
@@ -50,15 +51,15 @@ $colors: (
...
+    --color: #{map.get($variant, 'color')};
   }
 
   .bg-#{$key} {
-    background: map.get($variant, 'color') !important;
-    --background: map.get($variant, 'color') !important;
+    background: map.get($variant, 'color');
+    --background: map.get($variant, 'color');
     @if (map.get($variant, 'contrast') != null) {
-      color: map.get($variant, 'contrast') !important;
```
splashdust, line 61: Good riddance to those `!important` :D
This may be a breaking change for some consumers though, so we should mention it in the release notes.

splashdust (approved):

libs/chlorophyll/scss/components/utility/_colors.scss
splashdust: I think it's fine if we release it at the same time as #1045, which is also potentially breaking!

The use of `!important` here is probably a vestigial inheritance from Bootstrap, where utility classes usually had this to overcome specificity rules. Not sure if that is even needed here, but if it is, it would probably make more sense to use a cascade layer for that instead (i.e put all utility classes under `@layer utilities`, or something like that)

vsjolander: I'll leave that feature/fix for when we know more what chlorophyll will be in the future - just sticking to actual bug fixes for now. I feel that if I go down this rabbit hole I might not come up until a couple of weeks.... 🐰

Issue #1063, opened 2023-10-19
Checkbox: horizontal form-group overflow


Maximilian-R:
### Bug already reported?

- [X] I confirm that I have checked if the bug already has been reported

### For which framework/library you are reporting the bug

Chlorophyll

### Component name

form-group

### Description

Checkboxes in a horizontal layout should wrap - not overflow the parent container.

![image](https://github.com/sebgroup/green/assets/20689094/2e1cb3f1-0c62-4a6c-b5c2-53d7c1908d74)

### Current Behaviour

All checkboxes stays on the same row - no matter how wide it gets. 

### Expected Behaviour

Wrap checkboxes

splashdust:
Closed in #1064

PR #1064, opened 2023-10-19
fix(form-group): horizontal layout overflows

Maximilian-R:
#1063


vsjolander (approved):
Tested in firefox and can't find any problems with RadioGroup or Checkboxes.

PR #1070, opened 2023-10-24
docs(react/checkbox): updated docs with props and better examples

vsjolander:


libs/react/src/lib/form/checkbox/checkbox.stories.mdx
```diff
@@ -25,19 +38,15 @@ Simple Checkbox component
...
-    <Checkbox label="Valid unchecked" className="is-valid" />
-    <Checkbox label="Valid checked" className="is-valid" checked />
-  </form>
-  <form>
-    <Checkbox label="Invalid unchecked" className="is-invalid" />
-    <Checkbox label="Invalid checked" className="is-invalid" checked />
-  </form>
-  <form>
-    <Checkbox label="Disabled unchecked" disabled />
-    <Checkbox label="Disabled checked" disabled checked />
```
splashdust, line 40: @vsjolander I think we should keep `checked` in the docs. You still need to use it to programatically control the state, and ideally the component itself should be stateless (i.e the consuming app holds the state and sets the `checked` prop), so `defaultChecked` would probably be more of an edge case where you need to rely on the component's internal state for some reason.

splashdust (approved):

Issue #1072, opened 2023-10-25
Context Menu: Documentation


splashdust:
Several things are plain wrong in the Context menu documentation:

* The menu item click event is called `contextMenuItemClicked`, but referred to as `itemClicked` in an example
* The description of the above looks copy-pasted from the Slider component
* The import module name should be `NggContextMenuModule`

splashdust:
Closed in #1074

PR #1074, opened 2023-10-26
fix(context-menu): context menu imports and docs

### LINKED ISSUE #1071: Context Menu: Missing style dependencies
When context menu is imported from Chlorophyll, you don't get icon styles or popover styles which required by this component, and docs don't mention that you need to add them separately.

The correct behavior would be for these to be automatically imported by the context menu styles (i.e they should be `@use`d there)

splashdust:
Closes #1071 and #1072

Also adds some general notes on selective imports in Chlorophyll, and adds a missing dependency to Datepicker (it was missing icons import)


vsjolander (commented):
Looks good!

The only thing I'm wondering about is the context-menu in the angular-lib-dev, was this for testing perhaps? The component in the lib didn't change at all?

PR #1075, opened 2023-10-27
fix(alert-ribbon): don't render empty header

misael86:
Hi! Misael from team ToConsume on SEB.

Noticed that an extra margin was added to the alert ribbon when no header was passed, this due to an empty header div with style being added that had a margin right.

I propose to not render the header instead if it is empty. I have not been able to test this code so please try it out before merging it in.

Best regards Misael


splashdust:
@misael86 Thank you for the PR! Had to make a small tweak to the render logic, but otherwise all good!

vsjolander (approved):
Looks good! Thanks for contributing Misael! 💎 👏 😍

PR #1081, opened 2023-10-30
feat(react): new component InPageWizardStepCard

vsjolander:
### Needs to be fixed:
- Icon on edit button needs to be wrapped as a component and should be regular. It should also pick up the correct colour.


splashdust (approved):
Looking good! 👌 
Linter is failing, but it appears to be releated to the Tokens package rather than this PR

vsjolander:
Removed the linting rule since it was created by the nx generator and acted weired...

PR #1082, opened 2023-10-30
Green tokens: versioning and CI for tokens

vsjolander:
### What has changed?
- New file structure for tokens to match how they are categorised.
- project.json with targets to enable CI integration

### Comments
Not sure when publishing the first version of this package, should I do it manually or just do it through the action? What did you do when releasing core? @splashdust


splashdust:
> Not sure when publishing the first version of this package, should I do it manually or just do it through the action? What did you do when releasing core?

@vsjolander I did it through the action, but I ran the NX commands locally with the `--dry-run` flag first to weed out the simplest issues. Create a temporary branch first though, that way the clean-up gets easier if tags and release commits gets created accidentally.

vsjolander:
> @vsjolander I did it through the action, but I ran the NX commands locally with the `--dry-run` flag first to weed out the simplest issues. Create a temporary branch first though, that way the clean-up gets easier if tags and release commits gets created accidentally.

@splashdust Looks like you only can run `@jscutlery/semver:version` in the `main` branch since this is the `baseBranch`. So I wan't able to do it all from a temporary branch.. Maybe there was a way some how but I just did it all in `main` in the end.

PR #1085, opened 2023-10-31
fix: 🐛 slideout modal size prop

misael86:
Enable having different width on the slideout modal. Also added memo to all components and some accessibility area tags, same as the ones presented in the chlorophyl storybook


libs/react/README.md
```diff
@@ -42,7 +42,7 @@ yarn
 3. Start the Storybook
 
 ```bash
-yarn storybook react
+npx run react:storybook
```
misael86, line 45: This was the only way i managed to get storybook up and running, don't know if it is the correct way?

libs/react/src/lib/modal/modal.stories.tsx
```diff
@@ -67,13 +67,34 @@ ModalDefault.args = {
   dismiss: 'Nope',
 }
 
-export const SlideOut: Story<ModalProps> = Template.bind({})
-SlideOut.args = {
+export const SlideOutSmall: Story<ModalProps> = Template.bind({})
```
misael86, line 70: I don't know if this is the correct way of displaying different slide out versions?

libs/react/src/lib/modal/modal.tsx
```diff
@@ -1,5 +1,5 @@
 import { ModalType, Size } from '@sebgroup/extract'
-import { MouseEvent, ReactNode } from 'react'
+import { MouseEvent, ReactNode, memo } from 'react'
```
misael86, line 2: I've seen you guys don't use memo, how come?

libs/react/src/lib/modal/modal.tsx
```diff
@@ -66,14 +66,18 @@ const ModalFooter = ({
...
   const modalContent = () => {
     switch (type) {
       case 'slideout': {
+        let className: string | undefined = undefined;
+        if (size === "lg") className = 'gds-slide-out--960';
+        if (size === "md") className = 'gds-slide-out--768';
+
         return (
-          <aside role="dialog">
+          <aside role="dialog" className={className} aria-modal="true" aria-labelledby="modal_header" aria-describedby="modal_body">
```
misael86, line 80: added accessibility area props also, same as the ones used in the chlorophyl storybook

libs/react/README.md
splashdust: It depends a little bit on what you have installed on your machine, but yes I think `npx` makes more sense to show in the docs, since it often "just works" :)

The correct command should be `npx nx run react:storybook` though!

libs/react/src/lib/modal/modal.tsx
```diff
@@ -18,26 +18,26 @@ export interface ModalProps {
...
 
-const ModalHeader = ({ header = '', onClose }: Partial<ModalProps>) => {
+const ModalHeader = memo(({ header = '', onClose }: Partial<ModalProps>) => {
   const handleClose: ModalEventListener = (event) => {
     if (onClose) onClose(event)
   }
   return (
     <div className="header">
-      <h3>{header}</h3>
+      <h3 id="modal_header">{header}</h3>
```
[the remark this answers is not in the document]
splashdust: Yes, you should generate a uuid and store in a useState, so that every instance of the component gets a unique ID. The you can do ```id={`modal_header_${uuid}`}``` or similar.

Here's an example of this in another component: https://github.com/sebgroup/green/blob/main/libs/react/src/lib/accordion/accordion.tsx

libs/react/src/lib/modal/modal.stories.tsx
vsjolander: It's correct! However the whole documentation for modals could probably be improved.. But it's out of scope for this PR. Thanks for contributing <3

libs/react/README.md
misael86: Changed to npx nx run react:storybook :)

libs/react/src/lib/modal/modal.stories.tsx
misael86: Thank you for looking through the PR so fast!

libs/react/src/lib/modal/modal.tsx
misael86: From my own experience I've always used it as a best practice, even if some components don't need it or if the actual performance difference isn't that big. But if you start using React.memo it's also important to use React.useCallback, otherwise the component is still going to rerender every time due to the function passed to the component changing on every rerender.

https://stackoverflow.com/questions/53074551/when-should-you-not-use-react-memo

libs/react/src/lib/modal/modal.tsx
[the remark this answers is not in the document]
misael86: done! thanks for the example! helped a lot :)

libs/react/src/lib/modal/modal.tsx
splashdust: Thank you. I also looked around a bit more, and based on [this piece of advice](https://react.dev/reference/react/memo#should-you-add-memo-everywhere) in the React docs, I think we should skip using memo by default and only add it if some component has measurable performance issues.

Basically it has little to no benefit in most cases, but can potentially make troubleshooting other issues more difficult. Locally in an application, this is probably almost never an issue, but in Greens case I think we should favor ease of maintenance unless there are clear benefits to adding something.

misael86: Sure, the one importing the component can always wrap it in memo if needed. I've myself not experienced any problems when troubleshooting with memo, but I agree that the code becomes a bit more complex maybe. The times I've really needed it has been with huge lists of data. The only thing I feel is a bit ugly from what I've seen from SEB is that in some projects updating one input rerenders the entire page since the main state is loaded in the main component. Therefore I feel it's good to at least know about memo. But I removed it from this component now :)

misael86: Sure, the one importing the component can always wrap it in memo if needed. I've myself not experienced any problems when troubleshooting with memo, but I agree that the code becomes a bit more complex maybe. The times I've really needed it has been with huge lists of data. The only thing I feel is a bit ugly from what I've seen from SEB is that in some projects updating one input rerenders the entire page since the main state is loaded in the main component. Therefore I feel it's good to at least know about memo. But I removed it from this component now :)

splashdust (approved):
Looks good! 🥇
Thank you for the well-structured pull request! Explanations and motivations helps a lot in reviewing.

PR #1089, opened 2023-11-01
fix: 🐛 fix select border colour

misael86:
use same light gray border colour on select as on all other form elements


libs/chlorophyll/scss/components/select/_mixins.scss
```diff
@@ -7,12 +7,15 @@
   position: relative;
 
   select {
+    @include mixins.padding-horizontal(5);
```
misael86, line 10: Not sure the padding was the same as what it comes now, but it looks good in storybook, also this is the same setting that input has.

splashdust:
@misael86 Thank you for the pull request!

Actually, the border color of select was correct, but the other ones was incorrect :)
I updated the border color in the base config to the correct value, so now all inputs should get it.

libs/chlorophyll/scss/components/select/_mixins.scss
splashdust: The select needs additional right-hand padding to allow more space for the chevron, but otherwise it is correct. I added `padding-right` now, so you don't need to do anything.

splashdust (approved):
Thank you for another great PR! 🥇

PR #1091, opened 2023-11-02
feat(charts): add percentage formatting to tooltip

### LINKED ISSUE #687: Chart: Custom tooltip function for pie charts
I would like to be able to change the format for percentage presented in the tooltip for pie and donut charts. Like the tooltipNumberFormat function for bar charts.

image

Maximilian-R:
Align the tooltipNumberFormat function with the format function in billboard.js. 
Also allow it to override the formatting for charts with ratio. 

The current percentage is not always correct (adds a lot of decimals even for numbers that javascript can divide correctly. Seems like the ratio value is calculated by the angle of the barchart pies rather than the actual values in the underlaying implementation - superwierd). Therefor also limiting the default to 2 decimals

Closes #687

### DIFF
--- a/libs/charts/src/lib/billboard.spec.ts
+++ b/libs/charts/src/lib/billboard.spec.ts
@@ -263,6 +263,7 @@ describe('billboard', () => {
       const expected: Axis = {
         y: {
           show: false,
+          clipPath: false,
         },
       }
       expect(parsed.axis).toEqual(expected)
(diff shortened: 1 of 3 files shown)


comeab:
Hi @Maximilian-R 
Great job...Was working on a PR for this...Now when about to push i see your PR...Looks great
I was also adding a property to allow specifying the number of decimals for the percentage instead of toFixed(2)

Great job

splashdust (approved):
Looks really good, nice work as always! 💯

PR #1092, opened 2023-11-02
feat: 🎸 override tooltip number formatting

comeab:
Currently it is not possible to override the formatting of tooltips on the charts even with the existing properties

BREAKING CHANGE: 🧨 None

✅ Closes: #687

### DIFF
--- a/libs/charts/src/lib/billboard.spec.ts
+++ b/libs/charts/src/lib/billboard.spec.ts
@@ -324,13 +324,27 @@ describe('billboard', () => {
       )
       expect(formattedNumber).toEqual('100 kr')
     })
-    it('does not override tooltip percentages', () => {
+    it('override tooltip percentages', () => {
       const chartElement = '#foo'
       const settings: ChartSettings = {
         data: [],
         style: {
           tooltipNumberFormat: (value) => `${value} kr`,
         },
+      }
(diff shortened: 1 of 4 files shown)


splashdust:
Hey @comeab, thanks for the PR! By some cosmic coincidence, both you and @Maximilian-R submitted almost the same PR at the same time :)

I opted to approve @Maximilian-R's version mainly because I prefer to have a separate formatting function for percentages. Otherwise it could break formatting in existing implementations, since it would apply the same formatting to all values unless they update their formatting override.

I'm closing this one now and merging the other PR, but feel free to open a new one or get in touch if the other solution doesn't work for you!

comeab:
> Hey @comeab, thanks for the PR! By some cosmic coincidence, both you and @Maximilian-R submitted almost the same PR at the same time :)
> 
> I opted to approve @Maximilian-R's version mainly because I prefer to have a separate formatting function for percentages. Otherwise it could break formatting in existing implementations, since it would apply the same formatting to all values unless they update their formatting override.
> 
> I'm closing this one now and merging the other PR, but feel free to open a new one or get in touch if the other solution doesn't work for you!

hey no worries...i noticed @Maximilian-R  PR and actually commented on it. I love this collaborative work. 
Thats why i left this as a DRAFT...should have closed it actually


Thanks
Great work

PR #1093, opened 2023-11-02
Feat/add react textarea

misael86:
add text area component wrapped in form item with storybook stories and tests


.storybook/main.js
```diff
@@ -75,4 +75,8 @@ module.exports = {
     options.cache && (options.cache.set = () => Promise.resolve())
     return config
   },
+  typescript: {
```
misael86, line 78: enables listing all properties within storybook, did not list extended interfaces otherwise

libs/react/src/lib/form/textarea/textarea.spec.tsx
```diff
@@ -0,0 +1,112 @@
+/* eslint-disable jsx-a11y/no-redundant-roles */
```
misael86, line 1: tried to do similar tests to those on Input

libs/react/src/lib/form/textarea/textarea.tsx
```diff
@@ -0,0 +1,80 @@
+import { IExpandableInformation, ILabelAndLabelInformation, IValidator, randomId, validateClassName } from '@sebgroup/extract';
```
misael86, line 1: Tried to have the same structure as for input.tsx

libs/react/src/lib/form/textarea/textarea.tsx
```diff
@@ -0,0 +1,80 @@
...
+    role,
+    rows = 4,
+    style = {},
+    validator,
+    value,
+    "data-testid": dataTestId,
+    ...props
+  }: ITextAreaProps) => {
+    const [uuid] = useState(id);
+    const [localValue, setLocalValue] = useState(value);
```
misael86, line 34: I don't know how this works on input for example, but to be able to update what the user writes without getting a new value in through the props, I had to store a "temporary" value. For example in my team we only listen to the onBlur and update the state on that action. So the value props to TextArea is only updated on blur.

libs/react/src/lib/form/textarea/textarea.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+    const localOnChange = useCallback(
+      (event: ChangeEvent<HTMLTextAreaElement>) => {
+        setLocalValue(event.target.value);
+        if (onChange) onChange(event);
+      },
+      [setLocalValue, onChange]
+    );
+
+    const ariaDetails: string[] = [];
```
misael86, line 48: added some accessibility props, since both label info and expandable info can be react elements I read that it is better to have it as aria-details instead of aria-describedBy

libs/react/src/lib/form/textarea/textarea.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <div className="gds-input-wrapper">
+          <textarea
+            aria-details={ariaDetails.length > 0 ? ariaDetails.join(' ') : undefined}
+            autoComplete={autoComplete}
+            className={validator && validateClassName(validator.indicator)}
+            id={uuid}
+            onChange={localOnChange}
+            role={role}
+            rows={rows}
+            style={{ ...style, cursor: 'text', resize: 'none' }}
```
misael86, line 71: for some reason the default is a pointer cursor for text area, so I changed that and removed the option of resizing the input.

libs/react/src/lib/form/textarea/textarea.tsx
splashdust: Yes, this makes sense! I think this needs to be done for the other input fields as well, but that is outside the scope of this PR.

libs/react/src/lib/form/textarea/textarea.tsx
splashdust: I fixed the cursor thing in Chlorophyll instead. I suspect this is some old copy-past mistake that has stuck around :)

Textareas should still have the resize handle though, according to designs in Figma, so I added it back, but constrained to vertical. I also removed the `style` prop, because in essence, custom styling should not be supported in Green.

splashdust (approved):
Thanks for the PR, and really nice work once again! 💯

PR #1094, opened 2023-11-02
fix(react/dropdown): simplify how the React wrapper handles state

### LINKED ISSUE #1090: Dropdown: multiple selection support for controlled component
### Bug already reported?

React

### Component name

dropdown

### Description

When setting up a **multiple** selection `Dropdown` component in a controlled manner, the dropdown became unresponsive. From initial investigation, it seemed the code did not take into account the values being an array here https://github.com/sebgroup/green/blob/main/libs/react/src/lib/dropdown/dropdown.tsx#L102

```tsx
// example code
const [values, setValues] = useState([]);

return (
  <Dropdown
    label="a dropdown"
    multiSelect={true}
    options={[
      { label: 'aaa', value: 'aaa' },
      { label: 'bbb', value: 'bbb' },
    ]}
    value={values}
    onChange={setValues}
  />
);
```

### Steps To Reproduce

1. Click on multiple selection dropdown component
2. Click on any of the option
3. Notice non of the clicked options are being selected

### Current Behaviour

Dropdown is unresponsive when setup in a controlled manner.

### Expected Behaviour

We expect the multiple selection dropdown to be responsive when setup in a controlled manner.

splashdust:
The wrapper will now simply forward `value` to the underlying component, and value will correspond to the `[useValue]` field in the options object.

Closes #1090

### DIFF
--- a/libs/react/src/lib/dropdown/dropdown.stories.mdx
+++ b/libs/react/src/lib/dropdown/dropdown.stories.mdx
@@ -16,7 +16,7 @@ Dropdown Component
 <Story
   name="Dropdown"
   parameters={{
-    componentIds: ['component-dropdown']
+    componentIds: ['component-dropdown'],
   }}
   args={{
     loop: true,
@@ -36,7 +36,7 @@ Dropdown Component
     onChange: (value) => {
       console.log('OnChange', value)
     },
-    label: 'Dropdown label'
+    label: 'Dropdown label',
   }}
 >
   {Template.bind({})}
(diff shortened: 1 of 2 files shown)


eweseong (approved):

eweseong:
Hi @splashdust, the fix looked good to me :) is there a tentative date where the fixes would be release?

splashdust:
@eweseong I can publish a new release by the end of the day.
Thank you for the clear and easily reproducible issue description!

PR #1102, opened 2023-11-06
fix(react): in-page wizard edit button on mobile

vsjolander:
Edit button was not visible on mobile - it now is :)


splashdust (approved):
👍 👍

PR #1104, opened 2023-11-07
fix(angular/modal): 🐛 angular modal z-index

SEBNiklas:


splashdust (approved):
Hey @SEBNiklas, thanks for the PR!

It's always good to leave a few words about what problem the change solves, even for one-line fixes such as this one. It helps us understand what has changed, and why, more easily.

In this case the change is pretty obvious, so no biggie, but for next time! :)

Thanks again! 🙏

PR #1106, opened 2023-11-07
feat(localization): add possibility to add custom localization in Green Core

splashdust:
Add possibility to add custom localization in Green Core. The reasoning behind this is that we've removed the old `texts` input/props that components used to have to supply all default texts. In Green Core, the default texts are built-in and also have built-in localization. (these are texts such as the label for the input field in a searchable dropdown).

Typically you wouldn't need to override these built-in texts, but now it is possible if needed. Also, we'll probably use the same method to opt-in to additional locales that can be imported from Green Core. That way, if you only need Swedish and English, for example, you won't have to include any extra bloat from translations you don't use.


apps/core-lib-dev/src/app/custom-localization.element.ts
```diff
@@ -0,0 +1,75 @@
...
+// Hard coding localization templates like this is bad practice, but this
+// is just for testing.
+// DON'T DO IT!
+const extraLocales = ['fr']
+const extraTemplates = new Map([
+  [
+    'fr',
+    {
+      templates: {
+        s58bfb494feb8eb02: str`${0} baguettes`,
```
vsjolander, line 18: Best examples ever! 🤣

vsjolander (approved):
Looks good! When you have time I would love to go through the Lit localization part. How we should create and stor translations and stuff like that :)

Great work as always!

PR #1109, opened 2023-11-08
Tokens as Constructable Stylesheets

splashdust:
In order to avoid CSS variable leakage, we should not register tokens on `:root`, at least not by default. Instead we should use a shared stylesheet instance which loads the token variables into `:host` for each component.

This should enable all components to share the same variables without increasing bundle size or impacting performance, while also not leaking any variables outside of the shadow roots.


splashdust:
Merging this with the `vision` branch locally, because this now has bits of `main` in it 🤦

PR #1111, opened 2023-11-10
feat(react): groupFocus and className props

vsjolander:
Form groups that should have focus on the whole group now requires the `group-focus` class. There was no way to set this in the React component.

Now you can do this, as well as pass your own classes to the group if needed.


splashdust (approved):
👍 👍

PR #1115, opened 2023-11-14
feat: 🎸 fieldset react component

misael86:
Added a fieldset react component with padding, background and margin. Used to group inputs in a form.

image

Story uses https://github.com/sebgroup/green/pull/1119, so either approve that first or fix this one so it can be merged separately..


libs/react/src/lib/form/fieldset/fieldset.stories.mdx
```diff
@@ -0,0 +1,76 @@
...
+    <Input label="Input 1" />
+  </Fieldset>
+</Canvas>
+
+
+
+## Example
+
+<Canvas>
+  <div>
```
misael86, line 60: An example of the design we want to create

libs/react/src/lib/form/fieldset/fieldset.stories.mdx
```diff
@@ -0,0 +1,76 @@
...
+    <Fieldset marginBottom>
+      <Input label="Input 1" />
+    </Fieldset>
+    <Fieldset>
+      <Input label="Input 1" />
+    </Fieldset>
+  </div>
+</Canvas>
+
+## With legend
```
misael86, line 44: It's possible to add a legend, but I don't think it is part of the design system. But still wanted to show how it would look like.

splashdust:
Thank you for this contribution! I will mark this as draft for now, and get back when we've discussed in the Design System team where this component should live, how it's described, etc.

splashdust:
We have gone over this now an come to the following conclusions:

* Fieldset should not be tied to a specific style
* In the above example, the gray panel should actually be a Card, so there needs to be a grey variant of Card (we will add this)

So in the scenario in the screen shot, there would be a Card wrapping a Fieldset.

Thanks again for the PR @misael86! This type of feedback and detailed suggestions (in code) are instrumental for building a design system that works in the real world.

misael86:
@splashdust  Ok super! Would be nice if there's a way of adding or removing the margin bottom also. Sometimes when the gray box is in the end of a wizard step, then we do not want to add margin, but if we have more inputs below it, then the margin is nice. I also guess that the borders/shadow of the card will be removed when the gray version is used? 

You can throw away this PR if you don't need it any more. No problems for me :)

splashdust:
> Would be nice if there's a way of adding or removing the margin bottom also. Sometimes when the gray box is in the end of a wizard step, then we do not want to add margin, but if we have more inputs below it, then the margin is nice.

Good point, we'll take this into consideration!

> I also guess that the borders/shadow of the card will be removed when the gray version is used?

When we looked into this we also realised that the current Card needs some design adjustments, so I'm not sure what the outcome of that will be yet. Either it will keep the shadow even in this case, or there will be a shadowless variation.

astrit:
Thank you for your contribution @misael86 ! 

After reviewing the current features and the discussion we had on Teams, it seems that the functionality provided by the card and other layout components effectively covers the use case addressed in this PR. 

As agreed on teams I propose that we close this PR, as the new layout components fulfill the intended requirements. 

I appreciate your efforts and encourage you to explore the new features. If you have any questions or need assistance, please feel free to reach out.

Thanks again!

--- 

Feel free to modify any part of this comment to better fit your style or project guidelines!

PR #1116, opened 2023-11-14
fix: 🐛 wizard next icon margin left

misael86:
Added margin left to the next button icon, added some documentation, used classNames for className rendering, added stories that show how multiple wizard steps look beside each other and story showing possible versions of the step.

### DIFF
--- a/libs/react/src/lib/in-page-wizard/inPageWizardStepCard.stories.mdx
+++ b/libs/react/src/lib/in-page-wizard/inPageWizardStepCard.stories.mdx
@@ -1,5 +1,9 @@
-import { Meta, Story, Canvas } from '@storybook/addon-docs'
+import { ArgsTable, Meta, Story, Canvas } from '@storybook/addon-docs'
 import { InPageWizardStepCard } from './inPageWizardStepCard'
+import { Label, List, Value } from '../list/valueList.tsx'
+import { Input } from '../form/input/input.tsx'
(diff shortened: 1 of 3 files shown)


libs/react/src/lib/in-page-wizard/inPageWizardStepCard.stories.mdx
```diff
@@ -1,5 +1,9 @@
-import { Meta, Story, Canvas } from '@storybook/addon-docs'
+import { ArgsTable, Meta, Story, Canvas } from '@storybook/addon-docs'
```
misael86, line 1: Used ArgsTable plus documentation on props instead of handwriting the props table

splashdust:
Depends on #1119

Correct me if I'm wrong @misael86, but the Storybook example in this one depends on the updated Input from #1119?

Otherwise it looks good to me. And super nice that you managed to get the ArgsTable component working. That's definitely much better than manually writing those tables!

misael86:
That is correct @splashdust, I did all changes in one branch and then split them into different PRs, thought there might be stuff you don't like or need fixes in some and not in other. So that it will be easier to merge the changes and also review. But I forgot to change the story so that it uses the old inputs. But then I thought that if you liked the changes in the input PR then maybe I would not need to change this one. But let me know, I can update it otherwise so that it builds.

splashdust (approved):
Looks good to me, thank you @misael86 & @vsjolander!

PR #1117, opened 2023-11-14
feat: 🎸 Char counter to text area

misael86:
Added character counter to text area by adding prop to form item that permits adding text to form item footer

image
image


libs/react/src/lib/form/textarea/textarea.tsx
```diff
@@ -1,89 +1,113 @@
 import {
```
misael86, line 1: mostly formatting of the file, also added right aligned label to form item and used classnames for class string

libs/react/src/lib/formItem/formItem.tsx
```diff
@@ -140,6 +143,17 @@ export const FormItem = ({
       {children}
       <div className="gds-form-item__footer">
         {validator && <span className="form-info">{validator.message}</span>}
+        {rightAlignedFooterInfo && (
+          <span
+            className="form-info"
+            style={{
+              textAlign: 'right',
+              width: validator ? 'auto' : '100%',
```
misael86, line 151: this makes the text take the entire width when alone but only take the necessary width when displayed together with the validation error

splashdust (approved):
Looking good, thank you for the PR @misael86!

Issue #1060, opened 2023-10-17
Button: Missing class property


gustavlindstrom:
### Bug already reported?

- [X] I confirm that I have checked if the bug already has been reported

### For which framework/library you are reporting the bug

React

### Component name

Button

### Description

The react Button and Link with button prop component is missing the class property needed to set the danger class. This is availed on the chlorophyll [Button](https://sebgroup.github.io/green/latest/chlorophyll/?path=/docs/components-form-elements-button--button) but not on the react Button and Link component.

### Steps To Reproduce

Go to [Link](https://sebgroup.github.io/green/latest/react/?path=/docs/components-link--link) or [Button](https://sebgroup.github.io/green/latest/react/?path=/docs/components-form-button--button) and try to set the danger class to the component.

### Current Behaviour

Button and Link with button prop color is not able to be set to red/danger.

### Expected Behaviour

I expect the Button or Link with button prop to be of color red/danger.

Issue #824, opened 2023-04-01
Missing types on ButtonGroup like `className`


syradar:
### Bug already reported ?

- [X] I confirm that I have checked if the bug already has been reported

### For which framework/library you are reporting the bug

React

### Component name

ButtonGroup

### Description

It seems like the regular HTMLElement props are not being forwarded to the underlying element for ButtonGroup.

![image](https://user-images.githubusercontent.com/10453338/229313117-06680bef-0ef9-4717-bd82-ad2ba77603dc.png)

I got the error following the docs for ButtonGroup

<img width="742" alt="image" src="https://user-images.githubusercontent.com/10453338/229313304-07844516-e9b6-4fdc-83c5-692944a25b24.png">

### Steps To Reproduce

1. Use TypeScript
1. Add `className` to the ButtonGroup

### Current Behaviour

Complains that `className` is not a valid prop on ButtonGroup

### Expected Behaviour

It should either:

- Accept `className` and forward them to the underlying element.
- Accept a `buttonGroupSize` property that controls the size of the ButtonGroup.

PR #1118, opened 2023-11-14
fix: 🐛 expose props and pass them to elements

misael86:
Enable setting class on button, buttonGroup and link

✅ Closes: #1060, #824

### DIFF
--- a/libs/react/src/lib/form/button/button.tsx
+++ b/libs/react/src/lib/form/button/button.tsx
@@ -1,36 +1,34 @@
-import { ButtonHTMLAttributes, MouseEventHandler, ReactNode } from 'react'
-import { ButtonSize, ButtonType, ButtonVariant } from '@sebgroup/extract'
+import { ButtonHTMLAttributes, DetailedHTMLProps } from 'react'
+import { ButtonSize, ButtonVariant } from '@sebgroup/extract'
+import classNames from 'classnames'
 
-/* eslint-disable-next-line */
-export interface ButtonProps {
-  children?: ReactNode
-  type?: ButtonType
+export interface ButtonProps
+  extends DetailedHTMLProps<
+    ButtonHTMLAttributes<HTMLButtonElement>,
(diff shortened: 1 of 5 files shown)


libs/react/src/lib/form/button/button.tsx
```diff
@@ -1,36 +1,34 @@
...
+import { ButtonHTMLAttributes, DetailedHTMLProps } from 'react'
+import { ButtonSize, ButtonVariant } from '@sebgroup/extract'
+import classNames from 'classnames'
 
-/* eslint-disable-next-line */
-export interface ButtonProps {
-  children?: ReactNode
-  type?: ButtonType
+export interface ButtonProps
+  extends DetailedHTMLProps<
```
misael86, line 6: Pass all button props

libs/react/src/lib/form/button/button.tsx
```diff
@@ -1,36 +1,34 @@
...
-  size
+  size,
+  ...otherProps
 }: ButtonProps) {
-  const props: ButtonHTMLAttributes<HTMLButtonElement> = { type, disabled }
-  const classNames: string[] = []
-  if (variant) classNames.push(variant)
-  if (active) classNames.push('active')
-  if (size) classNames.push(size)
+  const buttonClassName =
```
misael86, line 26: use classnames instead

libs/react/src/lib/link/link.tsx
```diff
@@ -8,27 +9,23 @@ interface LinkProps extends HTMLProps<HTMLAnchorElement> {
 export const Link = ({
   button,
   children,
-  ...props
+  className,
+  role,
+  ...otherProps
 }: PropsWithChildren<LinkProps>) => {
-  const [anchorProps, setAnchorProps] = useState<HTMLProps<HTMLAnchorElement>>(
```
misael86, line 13: No need to store the props in state

Issue #910, opened 2023-06-02
Input: Add possibility to add symbol/text in right side of input field


DovileM:
### Does the component already exist

- [X] I confirm that I have checked if the component already exists in any of the Green libraries

### For which framework/library you are requesting the component

React

### Description

Missing feature to have symbol/text in right side of input field.
![image](https://github.com/sebgroup/green/assets/24279416/9e9b387d-dcb9-4f1a-b285-51351a1bc8fc)

See figma documentation of input field: 
- https://www.figma.com/file/dX0HJ2MOaQyAOTziCD5fQ7/SEB-Design-System?type=design&node-id=15098-59502&t=UWdlZzsrcyHwGLqw-0
- https://designlibrary.sebgroup.com/components/component-input

Issue #873, opened 2023-05-08
NumberInput: Support for showing a unit inside the field


daneriseb:
### Does the component already exist

- [X] I confirm that I have checked if the component already exists in any of the Green libraries

### For which framework/library you are requesting the component

React

### Description

Support for showing a unit inside the field.
Similar to https://github.com/sebgroup/green/issues/819 but for NumberInput.

vsjolander:
duplicate of #910

Issue #713, opened 2023-01-23
RadioGroup: Name property should be required?


daneriseb:
We noticed an issue in our MFE where we have multiple radio groups after each other. Pressing Tab didn't move you to the next radio group. We solved this by setting the name property for the radio groups.

To avoid this happening to other users of Green I suggest making the name property required.

Issue #569, opened 2022-09-05
Input: Display units


faridmasood:
NumberInput (React):
We require to display kr/mån and other help texts along with the input fields. In bootstrap, there are called add-ons (https://getbootstrap.com/docs/4.0/components/input-group/) and they can be added at the beginning or the end of the field.

The requirement looks like this on Figma
![image](https://user-images.githubusercontent.com/3430028/188405850-621b1d21-f1a4-4e72-8146-dec3d6fa1297.png)

PR #1119, opened 2023-11-14
feat: 🎸 Add unit to input and refactor

misael86:
DRAFT: Add unit to input and split checkbox and radiobutton. Some things do not work 100%. Disabled input does not show unit part as disabled, wrong background colour. Validation marker is not positioned correctly, Group ads margin bottom. I fixed it locally like this:

```
.gds-input-wrapper {
  &:before {
    z-index: 1;
  }

  .form-group > .group {
    margin-bottom: 0 !important;
  }
}
```

But I could not find how to do it in the green project.

BREAKING CHANGE: 🧨 RenderInput removed and changed for Input

✅ Closes: #926, #910, #873, #713, #713, #569

image


splashdust:
Summary of the changes in this PR:
* All text input fields (`type="text | email | number"`) are now implemented in a single component called `Input`
* `Radiobutton` and `Checkbox` are split as two separate components
* Inputs now support unit label
* Inputs now support custom formatting callback
* Inputs maintain local state now, but they still behave as controlled components when a value is supplied. There's been various discussions around this, and whether components should be "controlled" or not, but this approach seem to hit a good middle ground. A benefit is that the user can still input text even if the value prop is not refreshed. This simplifies use cases where you don't want to update app state on every `onChange`, for example if you only need to read the state on `onBlur`.

The old `RenderInput` is removed, but I don't think we need to consider this a breaking change. It was not mentioned in the docs, and I don't think it was ever intended to be public API. Also, tests pass fine without it, so I think we're good :)

Thank you for the PR @misael86!

PR #1120, opened 2023-11-15
Update button.ts

GauriKudtarkar:
Added 'link' option

### DIFF
--- a/libs/extract/src/lib/button.ts
+++ b/libs/extract/src/lib/button.ts
@@ -1,3 +1,3 @@
 export type ButtonType = 'button' | 'submit' | 'reset'
-export type ButtonVariant = 'primary' | 'secondary' | 'ghost' | 'tertiary' | 'link'
+export type ButtonVariant = 'primary' | 'secondary' | 'ghost' | 'tertiary' | 'link' | 'close'
 export type ButtonSize = 'small' | 'large'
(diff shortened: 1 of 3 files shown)


splashdust:
Hey @GauriKudtarkar, thank you for the PR!

Currently, a `link` variant of Button is not part of the design system, since it is preferred to use `<a href="...">` for links, and when you need a border-less button, there is the `tertiary` variant. Also, links should be used for navigation, while buttons are generally used for actions within the same page, so mixing these styles can be confusing for the user.

If the `tertiary` variant or regular anchor elements do not work for you, it would be good with some motivation. Then I can bring it up with the design system team to see if we need to revise anything regarding this.

GauriKudtarkar:
Hi, yes I think I can use link <a> tag , in our scenario we are using a link to open a slide out modal, so not really navigation as such. But I created this PR because I saw we already have a 'link' class in css, which we can use as variant. I was not aware of 'link' variant of button is not included in design system. Also where can I find what is desided in design system and what is not? Sorry I am yet to get  aquainted with all design system stuff.

splashdust:
@GauriKudtarkar I see! If it is an inline link in a text paragraph, then perhaps `<a role="button">` would work. Otherwise, the tertiary button would probably be best.
There used to be a link button in the design system, but it was removed/consolidated with the tertiary button.

image

To find out more about the design system and what components and variants are currently described, refer to Design Library: https://designlibrary.sebgroup.com/components

I will remove `link` before merging this, but we'll keep `close`.

Thanks again for the PR, and feel free to open more PRs or issues if you find more things that you are missing!

splashdust (approved):

PR #1132, opened 2023-11-28
fix: 🐛 use correct path for listbox-kb-nav-controller

BullishVince:
## Issue
When running [@sebgroup/green-angular@2.1.0](https://github.com/sebgroup/green/releases/tag/%40sebgroup%2Fgreen-angular%402.1.0), error appears "**Cannot find module 'src/controllers/listbox-kb-nav-controller' or its corresponding type declarations**"

## Solution
Use correct path for listbox-kb-nav-controller in listbox.ts and menu.ts

## Contributors
Team Comodo @ SEB

### DIFF
--- a/libs/core/src/primitives/listbox/listbox.ts
+++ b/libs/core/src/primitives/listbox/listbox.ts
@@ -14,7 +14,7 @@ import {
 import {
   ListboxKbNavController,
   ListboxKbNavigation,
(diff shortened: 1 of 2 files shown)


splashdust:
Oh, noticed this just now after already merging the same fix in #1133 :)

Will close this one, but thank you either way!

PR #1146, opened 2023-12-05
fix: 🐛 update track background when min and max values change

sjkayle:
update slider track background when input min and max has changed, so it follows the "thumb" value:
image


mario-subo (approved):

splashdust (approved):
Looks good to me. Thank you for the PR! 🥇

PR #1162, opened 2024-01-19
feat: 🎸Green react table component

terrance456:
- refactored bootstrap react component into using green.

1) Simple table

image

2) Sortable table

image

3) Selectable table

image

4) Expandable table

image


mario-subo (commented):
The look of the indeterminate checkbox does not seem correct, however it seems to be a chlorophyl issue with the checkbox itself so I would like to ask the rest of the maintainers if they want this this should be fixed as part of this PR or a separate PR? 


image

splashdust (approved):
Excellent work!

I have some concerns about screen reader support on expandable rows. I suspect they will not be properly announced as sub-rows to the parent. Also, the expand-button has no label, so it will be unclear what it does. I think this can be fixed in a follow-up PR though.

And I agree with @mario-subo that the intermediate checkbox styling issue is outside the scope of this PR.

I opened an issue for the checkbox thing: #1163

sjkayle (approved):

PR #1169, opened 2024-01-22
Fix: Slide-in modal transition

vsjolander:


libs/react/src/lib/modal/modal.tsx
```diff
@@ -89,6 +125,31 @@ export const Modal = ({
   ...props
 }: ModalProps) => {
   const [uuid, _] = useState(id)
+  const [status, setStatus] = useState<string>(UNMOUNTED)
+  const [shouldRender, setShouldRender] = useState<boolean | undefined>(false)
+
+  useEffect(() => {
+    console.log('useEffect', { isOpen, shouldRender, status })
```
splashdust, line 132: Stray console.log

splashdust (changes requested):
Found a stray `console.log`, otherwise it looks excellent!

Issue #1172, opened 2024-01-24
H5 and H6 should be set to medium font weight 


Fredrik-Oberg:
### Bug already reported?

- [X] I confirm that I have checked if the bug already has been reported

### For which framework/library you are reporting the bug

Chlorophyll

### Component name

Typography

### Description

H5 and H6 should be set to medium font weight according to Figma.

Figma reference: https://www.figma.com/file/dX0HJ2MOaQyAOTziCD5fQ7/SEB-Design-System?type=design&node-id=3-2&mode=design&t=xHZMlHL5h0DpPytq-0

Current code: 
https://github.com/sebgroup/green/blob/main/libs/chlorophyll/scss/tokens/_typography.scss#L173

PR #1175, opened 2024-01-25
fix: 🐛 Change font weight for H5 and H6 

fredrik-oberg-seb:
Font weight for H5 and H6 should be medium according to Figma spec
Figma: [SEB Design System - Font Styles](https://www.figma.com/file/dX0HJ2MOaQyAOTziCD5fQ7/SEB-Design-System?node-id=20783%3A58371)

image

✅ Closes: #1172


splashdust (approved):
Looks great. Thank you for the PR!

PR #1185, opened 2024-01-31
Ci: Switch to Changesets for versioning and publishing

splashdust:
This PR swaps out the current [semver](https://github.com/jscutlery/semver) setup with [Changesets](https://github.com/changesets/changesets).

## But why??
Semver relies on changelog information being stored in the commit messages. For Green, this has turned out to come with certain limitations. Since we are using *Squash and Merge*, there will only be one commit per PR on main, which is a big problem if a PR contains updates to multiple libraries. Of course, we could just use regular old merge commits and keep the full commit history on `main`, but this has two disadvantages:
- As maintainers, we have limited control over commit messages, which then translates to limited control over changelog content and versioning. We would have to rewrite commit history in PRs, which would be annoying for everyone.
- Commit history in main would become diluted with work-in-progress commits

## Ok, so how?
Changesets takes a different approach to encoding changelog information. Instead of relying on commit messages, this information gets stored explicitly in a markdown file, referred to as a *changeset*. You can read how this works in detail on the Changeset github repo, linked above.

In a nutshell, here's how it works:
```bash
$ npx changeset
```
This will bring up an interactive prompt where you can choose which packages should get bumped, and write a summary of the changes.

The resulting changeset file looks something like this:
```md
---
'@sebgroup/chlorophyll': minor
'@sebgroup/extract': patch
'@sebgroup/green-core': patch
---

Changes a bunch of things
```

This gives precise control over what gets released, and how the versions are bumped. It also allows us, as maintainers, to tweak this, if we need, before the actual release workflow runs.

When the PR gets merged to `main`, a release preparation workflow runs. This will open a release PR as an intermediate step. If more changes are merged to `main` before this PR is merged, it will get updated with the additional changes.

When we are happy with the accumulated changes in the release PR, we merge it to main, and that will cause the final release workflow to run and publish new package versions to npm.


astrit (approved):
Super great work Jocke, looks great, pretty good I guess finally having precise control over releases. 

One thing I do not understand though how does it store the version ?

splashdust:
> One thing I do not understand though how does it store the version ? 

The actual version numbers are stored in the `version` field of each package `package.json` file. This then gets updated when the changeset is applied in the versioning workflow.

PR #1193, opened 2024-01-31
fix(react,angular,core): updated style and docs

vsjolander:
## Todo
- aria-labels not working properly in react

### DIFF
--- a/libs/core/src/components/context-menu/context-menu.ts
+++ b/libs/core/src/components/context-menu/context-menu.ts
@@ -98,14 +98,14 @@ export class GdsContextMenu extends GdsElement {
       <button
         id="trigger"
         class="icon border-0 small ${classMap({ highlighted: this.open })}"
-        aria-label=${this.buttonLabel}
+        aria-label=${this.buttonLabel ?? this.label}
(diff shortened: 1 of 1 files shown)


vsjolander:
@splashdust I think this is a breaking change in React since we change the props.

Don't know if we should remap these in the wrapper but not sure how to do it the right way?

splashdust:
> @splashdust I think this is a breaking change in React since we change the props.
> 
> Don't know if we should remap these in the wrapper but not sure how to do it the right way?

That would require an additional wrapper layer I think. i.e wrap the Lit React wrapper in a second wrapper that remaps the props. That's what we do with the Dropdown and Datepicker.

But in this case, I wonder if we should not just keep the attribute as `label` after all, and just render the same value in both the button and the popover. Dropdown and Datepicker both use `label`, so this would also keep it consistent.

In other words, rename `buttonLabel` &rarr; `label`, remove `popoverLabel`, render `${this.label}` in both the button and the popover. Keep the `hideLabel` prop for hiding the text in the button. Then we version it as a minor change in all libs.

splashdust (approved):
Woho!!

PR #1203, opened 2024-02-07
feat(core): gds-menu-heading primitive

vsjolander:


splashdust (changes requested):
Looks super good! Just one issue related to custom element scoping (see review comment)

vsjolander:
@splashdust I needed to add gds-elements to the lookup table even if the scoping is turned off for your suggestion to work in both cases. Not sure if this was the right solution?

Also changed the name to `gds-option-heading` since it seemed more appropriate and added support in react and angular.

splashdust (changes requested):
Here's what we discussed: We rename `gds-option-heading` &rarr; `gds-menu-heading`, and allow it in both Context menu and Dropdown.

splashdust (approved):

PR #1210, opened 2024-02-09
add some input accessibility

misael86:


libs/extract/src/lib/helperFunction.ts
```diff
@@ -13,7 +13,7 @@ export interface IValidator {
   rules?: ValidatorRules
 }
 
-export const validateClassName = (value: IndicatorType) => {
+export const validateClassName = (value: IndicatorType | undefined) => {
```
misael86, line 16: Permit undefined so that it can be used with classNames easier

libs/react/src/lib/form/input/input.tsx
```diff
@@ -65,15 +66,13 @@ export const Input = ({
...
+  const handleChange = useCallback(
     (event: ChangeEvent<HTMLInputElement>) => {
       const oldValue = event.target.value
       const newValue = formatter ? formatter(oldValue) : oldValue
       setLocalValue(newValue)
 
-      event.target.value = newValue
-
-      if (onChange) onChange(event)
+      if (onChange) onChange({ ...event, target: { ...event.target, value: newValue } });
```
misael86, line 75: This is the only way I got formatting to work so that the caret doesn't jump to the end when removing text in the middle of the field. For some reason doing event.target.value = newValue does not work.

splashdust:
@misael86 Thank you for the PR!

Some of the React tests are failing, and I see that some interfaces have changed, which looks like it might include some breaking changes. If these are necessary breaking changes, it would be good with a summary of why it is needed. That would also help us review this PR faster.

libs/react/src/lib/form/checkbox/checkbox.tsx
```diff
@@ -17,36 +16,50 @@ export interface CheckboxProps extends HTMLProps<HTMLInputElement> {
...
   label,
   onChange,
   validator,
   testId,
+  required,
+  className,
   ...props
 }: CheckboxProps) => {
-  const inputProps = useInput(props, onChange)
+  const [uuid] = useState(id)
```
misael86, line 29: only set an id once, not every time the component is rendered

libs/react/src/lib/select/select.tsx
```diff
@@ -1,121 +1,98 @@
...
-  autofocus?: boolean
-  form?: string
-  name?: string
-  required?: boolean
-  size?: number
-  selected?: boolean
-}
-
-interface OptionGroupProps {
+export interface SelectProps
```
misael86, line 13: use the react props definitions for select instead of defining everything manually

libs/react/src/lib/select/select.tsx
```diff
@@ -1,121 +1,98 @@
...
+  extends IExpandableInformation,
+    ILabelAndLabelInformation,
+    React.DetailedHTMLProps<
+      React.SelectHTMLAttributes<HTMLSelectElement>, 
+      HTMLSelectElement
+    > {
   label: string
-  children: React.ReactElement<OptionProps> | React.ReactElement<OptionProps>[]
-  disabled?: boolean
+  info?: string
```
misael86, line 21: this was renamed to info from expandable info since all other components have that name. This is a breaking change and can be reverted, not a necessary part of the PR

libs/react/src/lib/select/select.tsx
```diff
@@ -1,121 +1,98 @@
...
       expandableInfo,
       expandableInfoButtonLabel,
-      name,
+      required,
+      ...props
     },
     ref
   ) => {
-    const selectId = id ?? randomId()
+    const [uuid] = useState(id)
```
misael86, line 43: set id once and not every time the component is rerendered

libs/react/src/lib/select/select.tsx
```diff
@@ -1,121 +1,98 @@
...
+          />
           <ChevronDown />
         </div>
       </FormItem>
     )
   }
 )
 
-export const Option = ({ value, children, ...rest }: OptionProps) => {
+type OptionProps = React.DetailedHTMLProps<React.OptionHTMLAttributes<HTMLOptionElement>, HTMLOptionElement>;
```
misael86, line 82: use these instead of defining manually

libs/react/src/lib/select/select.tsx
```diff
@@ -1,121 +1,98 @@
...
       >
-        <div className={`gds-select ${validatorClassName}`}>
+        <div className={selectClassName}>
           <select
-            id={selectId}
+            aria-describedby={describedBy || undefined}
+            aria-invalid={validator?.indicator === 'error'}
+            aria-required={required}
+            id={uuid}
             data-testid={testId}
```
misael86, line 70: removed props are passed with the ...props instead, no need to do as before

libs/react/src/lib/form/input/input.tsx
```diff
@@ -65,56 +66,62 @@ export const Input = ({
     setLocalValue(value)
   }, [value])
 
-  const localOnChange = useCallback(
+  const handleChange = useCallback(
     (event: ChangeEvent<HTMLInputElement>) => {
```
misael86, line 70: Updated how the caret position is selected, did not work as it was before. Now we always keep the same caret position as earlier, but if the formatting ads or removes chars, then we move the caret the same number of steps as chars added. I have tested it with two of our formatters, might not work for all though.

Formatters:
formatAmount 
formatPersonalIdentificationNumber 
https://github.sebank.se/toconsume/consumer-loan-mfe/blob/develop/src/common/helpers/formatters.ts

misael86:
@splashdust I've added some more comments and also fixed the broken tests. If there are any other questions, please let me know! As I said in one comment, I've tested the caret positioning with our formatters, but I can nor promise that there won't be any other formatters that might need some extra code.

splashdust:
@misael86 Excellent stuff, these extra comments are super helpful! Will hopefully be able to digest all of it and review during next week. Cheers!

misael86:
@splashdust Reach out to me on Teams otherwise :) Misael Berrios Salas

splashdust (approved):
Looking good, thanks for contributing! 🥇

PR #1232, opened 2024-02-27
feat: Grouped list 

vsjolander:
This pull request introduces a new feature in our library, the `Grouped list` component. This component is designed to display many datapoints with labels in a structured way. The datapoints can be grouped under headings and may include a link or button to act on the data being displayed.

This PR contains:
- A `Green Core` component and documentation for it
- A `Green React` component derived from the lit component in core
- Documentation for how to use the Green Core component in Angular in the `Green Angular` documentation


splashdust (approved):
Changed `li` -> `gds-list-item` as we discussed. Also made a component of it after all. Not much overhead, and then it can add `role` and `gds-element` by itself.

PR #1239, opened 2024-03-01
Datepicker: Add support for disabling dates

### LINKED ISSUE #615: Datepicker: Enable Bank days only
The requirement is to disable the non bank days (Official Vacations and Weekends ... etc.)
But Date Picker Component doesn't have the ability to disable some days.

splashdust:
**Datepicker:** Added support for disabling dates

- `disabled-weekends` will disable all weekend days in the calendar view
- `disabled-dates` allows to specify an array of dates that should be disabled in the calendar view

Closes #615


vsjolander (approved):
Awesome as always!

Issue #1225, opened 2024-02-22
Dropdown: Not selecting the values null, 0, ""


draganseb:
### Bug already reported?

- [X] I confirm that I have checked if the bug already has been reported

### For which framework/library you are reporting the bug

React

### Component name

dropdown

### Description

We are having issues with the Dropdown component using Green React. When the value is null, undefined, 0, "" it doesn't select that value and ignores the change. I think that this values are legit and should be accepted. 
 
This is the code where the value is first checked and then applied. 
https://github.com/sebgroup/green/blob/16ecaad5ff02a399d150a054a385e5352895bc70/libs/react/src/lib/dropdown/dropdown.tsx#L100

### Current Behaviour

When the value is null, 0, "" it doesn't select that value and ignores the change. 

### Expected Behaviour

The values null, 0, "" should be allowed values and the dropdown should apply them

PR #1240, opened 2024-03-01
fix: 🐛 the dropdown in react should select falsy values

draganseb:
the dropdown in react should select falsy values like "", 0, null, undefined instead of ignoring them

✅ Closes: #1225

### DIFF
--- a/libs/react/src/lib/dropdown/dropdown.tsx
+++ b/libs/react/src/lib/dropdown/dropdown.tsx
@@ -96,8 +96,8 @@ export const Dropdown = ({
   ...props
 }: DropdownProps) => {
   const handleOnChange = (e: any) => {
-    if (e.detail?.value) {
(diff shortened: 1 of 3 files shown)


splashdust (approved):
Looks good to me! Thank you for submitting a fix for this issue! 🥇

libs/react/src/lib/dropdown/dropdown.tsx
```diff
@@ -96,8 +96,8 @@ export const Dropdown = ({
   ...props
 }: DropdownProps) => {
   const handleOnChange = (e: any) => {
-    if (e.detail?.value) {
-      onChange?.(e.detail.value)
+    if ('value' in e.detail) {
```
draganseb, line 99: Will just check if the 'value' property exists in details object

PR #1251, opened 2024-03-08
Add ValueList functionality

misael86:
Added different styles based on the design library.

image


splashdust:
Hey @misael86, since [Grouped List](https://sebgroup.github.io/green/latest/react/?path=/docs/components-grouped-list--page) have been added to Green Core, we've come to the conclusion that the React version should be deprecated and not receive further updates. I've opened a new PR (#1283) to reflect this change.

With that I will close this one, but nevertheless, big thanks for contributing to Green. PRs like this are always highly appreciated! 💚

PR #1257, opened 2024-03-14
fix: tokens now builds correctly after install

vsjolander:
By setting matching versions in package.json we don't get a local node_modules in the `tokens` package.


splashdust:
@vsjolander I think these dependencies should not be in this packages json at all though, since they are not run-time dependencies of the package.
They are only used when building, so they only need to be dependencies of the mono-repo.

vsjolander:
@splashdust yes you are correct. 

I think my idea was that when installing the package you could use it to build your own required output but since I don't include the src that's not possible :D

Removed the dependencies all together from the lib package.json.

splashdust (approved):
💯

PR #1258, opened 2024-03-14
Grid V1.0 & Tokens

astrit:
# Grid

The `gds-grid` is a custom element that provides a flexible grid system. It uses CSS grid layout to arrange its child elements into columns. This component is highly customisable and responsive, allowing you to specify the number of `columns`, `gap`, `padding`, and fluidity for different devices with automated column size based on the content using the `fluid` attribute.

### Key Features
-  **Customizable Grid Layout:** Utilize the `gds-grid` element to create grids with specified columns for desktop, tablet, and mobile.
-  **Responsive Design:** Tailor column numbers, gaps, and padding for each device to ensure optimal display across different screen sizes.
-  **Fluidity Option:** Enable the `fluid` attribute to allow the grid to adjust column sizes based on content and available width.
-  **Internal Design System:** Incorporate predefined size references for gaps and padding to maintain consistency in design.

### Next Steps
-  Gather user feedback on the new grid features for further refinements.
-  Explore potential enhancements such as additional layout options or dynamic resizing capabilities based on device orientation and potentially use `grid-area` for specific layouts.


libs/core/src/components/grid/grid.ts
```diff
@@ -0,0 +1,126 @@
...
+        text-wrap: balance;
+      `
+    }
+
+    const sheet = new CSSStyleSheet()
+    sheet.replaceSync(`:host {${cssVariables}}`)
+
+    if (this.shadowRoot) {
+      this.shadowRoot.adoptedStyleSheets = [
+        ...this.shadowRoot.adoptedStyleSheets,
```
splashdust, line 117: I suspect that this will cause a new set of css variables to be added each time a property is changed.

Since you create a new stylesheet on lines 112..113 and then add it together with existing stylesheets here, it will keep adding new copies on each update (though I haven't verified this in practice)

A better approach may be to have a state variable with the set of grid variables, then update that state variable when one of those properties changes, and then add a watcher on the state variable to update the `adoptedStyleSheets` by combining with the static `styles` from the class.

Something along these lines:

```ts
// State variable to keep track of the grid variables
@state()
private _gridVariables = {
  column: css`...`,
  gap: css`...`,
  ...etc
}

// Watch the grid properties and update state variable when something changes
@watch('columns')
private _handleColumnsChanged() {
  this._gridVariables = {
    ...this._gridVariables,
    column: css`...new column variabels...`
  }
}
// ...etc for the rest of the properties

// Watch the state variable and update `adoptedStyleSheets` when it changes
@watch('_gridVariables')
private _updateGridCss() {
  const sheet = new CSSStyleSheet()
  sheet.replaceSync(`:host {${Object.values(this._gridVariables).join('')}}`) // join the state object to a string
  
  if (this.shadowRoot) {
    this.shadowRoot.adoptedStyleSheets = [
      ...GdsGrid.styles, // This should get the static styles added to the class, rather than the accumulated set of stylesheets
      sheet
  }
}
```

libs/core/src/components/grid/grid.ts
```diff
@@ -0,0 +1,126 @@
...
+  @property({ attribute: 'gap', type: String })
+  gap?: GridGap
+
+  @property({ attribute: 'padding', type: String })
+  padding?: GridPadding
+
+  @property({ attribute: 'fluid', type: String })
+  fluid?: boolean
+
+  updated(changedProperties: Map<string | number | symbol, unknown>) {
```
splashdust, line 37: Instead of using `updated` and checking for each attribute name here, you could also use the `@watch` decorator.

Example:
```ts
@watch('columns')
private _handleColumnsChanged() {
  // ... code for columns changed
}
```

libs/tokens/src/formats.ts
```diff
@@ -40,6 +40,77 @@ const formats: Record<string, Format> = {
...
+        formatHelpers.formattedVariables({
+          format: 'css',
+          dictionary,
+          outputReferences: options.outputReferences,
+        }) +
+        `\n}\n`
+      )
+    },
+  },
+  // size: {
```
splashdust, line 85: No need to keep commented out code around. Just nuke it 💥 We can get it from git history if necessary :)

libs/core/src/components/grid/grid.react.ts
splashdust, line 1: You can remove this one. React wrappers can be added in the React library instead (that way we don't need Core to depend on React)

splashdust (changes requested):
Over all this looks super good to me! Really amazing work🔥

A few minor things caught my review eye. I added inline comments for those. Mostly tiny things, but the one about `@watch` and state variables may need some clarification. We can have a look at this in person if you like!

libs/tokens/src/formats.ts
astrit: Good catch, actually forgot about it 🤘

libs/core/src/components/grid/grid.react.ts
astrit: This sounds good, I did remove it, good to know that is not necessary in the future on other comp...

libs/core/src/components/grid/grid.ts
astrit: Updating the stylesheet not sure if this is the best way to do it but please check the update and let me know 🤘
It could potentially do with some further refactoring reducing repetition for updating props, suggestions are welcomed ✌️

libs/tokens/src/tokens/space/margins.json
```diff
@@ -0,0 +1,128 @@
+{
+    "sys": {
+        "space": {
+            "margins": {
+                "0": {
```
vsjolander, line 5: Should these also be "xs, s, m, l....." maybe?

However they are many 😄 OK as is but lets think about it 👍

libs/tokens/src/tokens/space/padding.json
```diff
@@ -0,0 +1,128 @@
+{
+  "sys": {
+    "space": {
+      "padding": {
+        "0": {
```
vsjolander, line 5: Same here with the size scale

vsjolander (commented):
Great work!! Some minor stuff with the largest thing being the columns attribute syntax.

Interested to hear your thoughts on it @astrit 👊

libs/core/src/components/grid/grid.ts
```diff
@@ -0,0 +1,126 @@
...
+  @property({ attribute: 'columns', type: String })
+  columns?: string
+
+  @property({ attribute: 'gap', type: String })
+  gap?: GridGap
+
+  @property({ attribute: 'padding', type: String })
+  padding?: GridPadding
+
+  @property({ attribute: 'fluid', type: String })
```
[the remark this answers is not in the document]
astrit: Excellent, sounds good to me also, `auto-columns` makes sense. 
`columns="auto"` would be even more efficient but It might make it difficult to understand what is going on. 

Perhaps we could use same concept here if you want on `auto-columns` different sizes on different breakpoints. 
Will explore that option.

libs/core/src/components/grid/grid.stories.mdx
```diff
@@ -0,0 +1,111 @@
...
+    <!-- Child elements here -->
+</gds-grid>
+```
+In the example above, the grid will have `8` columns on `desktop`, `4` on `tablet`, and `2` on `mobile`.
+
+## Columns
+
+The `columns` attribute is used to define the number of columns for different devices.
+
+The format is `columns="[DESKTOP] [TABLET] [MOBILE]"`.
```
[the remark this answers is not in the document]
astrit: In the upcoming version, I intended to implement a more comprehensive method for managing breakpoints. However, after considering the matter further, I believe that placing the breakpoint beforehand contributes to clarity will go with it ✌️

I had thought of an additional attribute like breakpoints, allowing for the specification of widths like ﻿breakpoints="1200px 600px 400px" but I guess this won't be necessary since it doesn't really change that often, if it is requested on the future would be easy to extend. 

Will update the attribute accordingly 👌

libs/tokens/src/tokens/space/margins.json
astrit: I think the gap and paddings on grid it self make sense since is limited to certain amounts, but in here since the list is extensive I think is going to be difficult to name. But definitely we can look into it ✌️

libs/core/src/components/grid/grid.stories.mdx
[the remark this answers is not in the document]
vsjolander: Me and Jocke talked about this aswell! It will be messy how to handle/generate the mediaqueries for this, you end up recreating a middle layer for CSS when it's probably just easier to just use CSS in your app to do this configuration. And if you have special requirement for overriding the default stying of this grid component, you are probably better off writing the styling yourself anyway :)

libs/core/src/components/grid/grid.ts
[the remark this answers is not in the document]
vsjolander: Yea that's a great idea! 💎  We didn't even think about this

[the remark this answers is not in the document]
astrit: This is updated now! 
`auto-columns` supports like other attributes multi values or single for all breakpoints.

libs/core/src/components/grid/grid.stories.mdx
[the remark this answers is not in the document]
astrit: This is now updated

astrit:
If the updates seem complex to follow even though I explain as best as I could especially the difference with and without `auto-columns` and now it introduces single values as fallback for all breakpoints or the format l: for desktop, m: for tablet, and s: for mobile on all attributes. 

The formula is like this: `MaxColumnWidth = (W - ((C - 1) * G)) / C`, where `W` is the total available width, `C` is the number of columns, and `G` is the width of a single gap. This calculated width is then used to set each column's width, with a minimum width of the `auto-columns` value.

This is optional now and can be used without breakpoints at all which makes it even better or combined with columns. 

I did also merge the `feat/tokens-2016` branch. 

@splashdust & @vsjolander please check it again and let me know.

splashdust (approved):
Excellent work! 👌

PR #1277, opened 2024-03-22
Segmented Control

splashdust:
This PR adds the Segmented Control to Green Core.

### Design
This component supports both 2023 styles and transitional styles for the 2016 design.

### Responsivity
Layout is handled by a custom algorithm controlled by a ResizeObserver, in order to achieve a responsive layout that can be both scrolled and animated.

### Accessibility
The segmented control is presented as a list where each segment is a list item. The currently selected item is marked with `aria-current`. The segments can be navigated using regular tab and shift-tab. When focus reaches a segment that is out of view, it will automatically scroll into view.

### API
The component exposes a simple event based API, where each segment can be assigned a value and the value of the currently selected segment will be reflected in the `value` property of the parent component. A `change` event is dispatched when the value is changed by the user.


astrit (approved):
Looking fantastic! I tested it out and really like it.

One minor suggestion I have is regarding the scrolling behaviour. I found myself naturally wanting to scroll horizontally using a trackpad. Would enabling horizontal scroll with a `hidden` scrollbar and `﻿scroll-snap` enhance the overall navigation experience?

splashdust:
@astrit Scroll snap is a really nice idea! If that works we could probably cut down a 100 lines of this code :)

But for now, let's merge it as is, and then it can be improved in a separate PR.

PR #1280, opened 2024-03-27
Stylelinting updates and fixes

vsjolander:
## Done in this PR 
- Fixed stylelinting issues
- Upgraded styellinting packages
- Added custom-property-pattern


libs/core/src/components/datepicker/datepicker.trans.styles.scss
```diff
@@ -37,24 +37,24 @@
     background-color: var(--sg-form-control-bg);
     box-sizing: border-box;
     color: var(--sg-text-primary);
+    cursor: text;
     display: flex;
     justify-content: space-between;
-    align-items: stretch;
```
splashdust, line 42: This breaks the focus outline:
Screenshot 2024-03-28 at 08 29 10

libs/core/src/components/datepicker/datepicker.trans.styles.scss
```diff
@@ -37,24 +37,24 @@
...
     margin: 0.5rem 0;
     min-height: 2.75rem;
-    cursor: text;
 
     &:hover {
-      background-color: var(--grey-200);
+      background-color: var(--_grey-200);
```
splashdust, line 47: This variable does not exist, which breaks the hover style

splashdust (changes requested):
Found two (most likely unintentional) things. Can you have a look?

PR #1289, opened 2024-04-03
feat(chlorophyll): custom use-scope selector

Maximilian-R:


splashdust (approved):
Great idea! Looks good to me. 💯

Issue #1298, opened 2024-04-08
Grid issues


daggelito02:
Hi Astrit,

I found some issues in you fabulous grid.

1. There are no gap space in the horizontal (row) line. (see picture)
2. Specifying different values for the gap (Example: gap="l:xl m:l s:xs") doesn't have any effect. Alvays 32px for all view ports.

Bonus feedback. https://sebgroup.github.io/green/latest/core/?path=/docs/components-grid--page 
The values on columns should have the size/media value besides the amount of columns. 
![image](https://github.com/sebgroup/green/assets/40634037/6de08b31-294d-465b-87db-c65a15bafa30)

<gds-grid columns=**"8 4 2**" auto-columns="l:200 m:100 s:80">
    
</gds-grid>
(columns="**l**:8 **m**:4 **s**:2").

![image](https://github.com/sebgroup/green/assets/40634037/e2a815d8-6c62-45c3-8976-7c0d03fad9d3)

@astrit

astrit:
Hi @daggelito02 

Super nice that you already started to use Grid, and happy to know you like it. 

Now to answer your questions: 
1. There isn't at the moment a functionality to add row gap, we might introduce that feature in the future this could be a good test, how ever it is meant to do that by adding `margin-block-end` to the child it self at least for now. 
2. That is a bit odd on my setup it works and I did test it out again to make sure but we can take a closer look at your setup and try to debug.  

Here is an example on the `core-lib-dev`
![CleanShot 2024-04-08 at 23 23 52@2x](https://github.com/sebgroup/green/assets/2398447/a7e7ed73-1b87-4802-b99a-fd2f312cec37)

Good catch on the last one as that is a typo actually only on the documentation as in fact still needs the device prefix. 

Let me know when we can check on this!

🤘

astrit:
@daggelito02 these issues were addressed on PR: #1302 , which has been published. Please upgrade `core` to `1.12.0`

The release:
https://github.com/sebgroup/green/releases/tag/%40sebgroup%2Fgreen-core%401.12.0

PR #1299, opened 2024-04-08
Fix: Typo Grid — #1298

astrit:
Very minor change a mistake on the grid example when using auto columns and columns combined.

Good catch @daggelito02


splashdust (approved):
👌

PR #1310, opened 2024-04-12
Component: Button 

vsjolander:
This is PR for finalizing the button component in LIT for 2016 and 2023 design.

## Todo:

In order to publish the button the button in our code packages we need to:

- [x] Add correct tokens
- [x] Add exports of components
- [x] Add the stories back
- [x] Transitional styles
- [x] Component attributes (Sets, variant and hirarchy)
- [x] We decided that the current attributes that are called variant and sets are incorrect.

**After some discussion in the dev team we decideded the following:**

variant attribute should be used but for changing the appearance/color of the component (button)
remove sets attribute
add rank to controll primary, secondary, tertiary button styles
We can't use style, role since they are reserved in HTML spec.


vsjolander:
Moved work to #1329

PR #1311, opened 2024-04-12
Add `gds-theme` component

splashdust:
## `gds-theme`

This PR adds a new component called `gds-theme`. This component is responsible for supplying its descendant DOM with CSS variables, and can be used to toggle dark and light mode for that part of the DOM tree.

This initial implementation can be considered as an optional beta. No components rely on this yet, but it can be used to supply 2016 variables to components when using `transitionalStyles` without also having Chlorophyll loaded in the document.

It will work like this:

```html
<gds-theme color-scheme='dark'>
  <!-- Everything under here will use dark mode regardless of device setting -->
</gds-theme>

<gds-theme color-scheme='light'>
  <!-- Everything under here will use light mode regardless of device setting -->
</gds-theme>

<gds-theme color-scheme='auto'>
  <!-- Everything under here will use device setting -->
  <!-- Omitting [color-scheme] will default to 'auto' -->
</gds-theme>

<!-- Themes can also be nested -->
<gds-theme color-scheme='light'>
  <gds-card>  
    <h2>This is a light card</h2>
    <p>But sometimes, parts of the page need to use a different mode.</p>
    <gds-theme color-scheme='dark'>
      <gds-card>
        <h3>A dark card within!</h3>
        <p>This will use dark-mode, even though the rest of the page is light-mode</p>
      </gds-card>
    </gds-theme>
  </gds-card>
</gds-theme>
```

Potentially, in a future update, we can also move the `transitionalStyles` feature here in order to make it declarative and allow for more fine-grained control over which components use the 2016 styling.

Along the lines of this:
```html
<gds-theme design="2016">
  <!-- Everything here will use 2016 styles -->
</gds-theme>
```


astrit (approved):
Quite an update, lot's of components, looks good to me ✌️

splashdust:
@astrit Awesome! I just discovered some issues related to dark mode that I need to solve before merging 👍

PR #1316, opened 2024-04-16
Merge/docs

astrit:
This PR merges docs ( seb.io ) into main Green repo. 
Also sets the workflows to publish on staging and production.


apps/docs/public/fonts/SEBSansSerif-Bold.woff
splashdust, line 1: The font files should come from the fonts package, right? Shouldn't need to bundle them here

apps/docs/public/core-out/index.js
splashdust, line 1: This one should be removed

splashdust (changes requested):

apps/docs/public/fonts/SEBSansSerif-Bold.woff
astrit: I guess will update this once we move the fonts package in here also ✌️

splashdust (approved):

vsjolander (approved):
LGTM!

PR #1317, opened 2024-04-17
Fix `gds-icon` element scoping

splashdust:
`gds-icon` was using the wrong custom element decorator, causing it to register unscoped.


vsjolander (approved):
LGTM!

PR #1321, opened 2024-04-18
Migrate Fonts

astrit:
Initiated `@sebgroup/fonts` migration


vsjolander (changes requested):
Looks good! Some things I found:

1. The build step seems to fail if the dist folder structure is not created before hand:
Screenshot 2024-04-19 at 14 59 38

2. You will need the [package.json](https://github.com/sebgroup/fonts/blob/master/package.json) aswell and copy it over to dist folder so it can be distributed as a package.

astrit:
@vsjolander @splashdust 

The package is updated now, what it does: 
1. Copy fonts 
2. Build CSS 
3. Copy package.json
4. Copy README

Let me know if anything else should be done.

astrit (commented):
On first step now making sure to always create the fonts dir

astrit:
@vsjolander or @splashdust I am going to need your assistance here to configure the paths and how they get resolved because when building `chlorophyll` fails to properly resolve paths!

vsjolander:
Something caused react-charts tests to fail. Moving to a cleaner branch: #1377

PR #1323, opened 2024-04-22
Add property for columns to fix init count on safari

astrit:
This is a minor fix for Safari where it would not get properly the initial count as zero and then it caused the auto columns to stop working entirely!


splashdust (approved):
Looking good, nice work!

PR #1328, opened 2024-04-23
Chlorophyll: Update card border radius

splashdust:
Update card border radius to match Figma


vsjolander (approved):
Checked with Figma and it aligns!

PR #1329, opened 2024-04-23
New component: Button

vsjolander:
This is PR for finalizing the button component in LIT for 2016 and 2023 design.

## Todo:

In order to publish the button the button in our code packages we need to:

- [x] Add correct tokens
- [x] Add exports of components
- [x] Add the stories back
- [x] Transitional styles
- [x] Component attributes (Sets, variant and hirarchy)
- [x] We decided that the current attributes that are called variant and sets are incorrect.

**After some discussion in the dev team we decideded the following:**

variant attribute should be used but for changing the appearance/color of the component (button)
remove sets attribute
add rank to controll primary, secondary, tertiary button styles
We can't use style, role since they are reserved in HTML spec.


astrit (approved):
Looks good to me, super nice to see the 2016 style playing along the 2023. 
Hopefully soon we can update the font also to fix the `line-height` ✌️

splashdust (approved):
Looks great! So nice to finally publish the button in Core!

splashdust (changes requested):
Just realized one thing. The tests are a bit rudimentary still.

These are missing:
- [ ] Axe smoketest (to.be.accessible)
- [ ] Icon button
- [x] Rank attribute and property
- [x] Variant attribrute and property

splashdust (approved):
🧪 👍

PR #1333, opened 2024-04-24
Lib Tokens copy readme

astrit:
I did create a new command to copy readme separately as yo use the assets apparently you should be using one of the executors that supports `assets` option.


splashdust (approved):
👍

PR #1344, opened 2024-04-26
Fix Accordion dark mode

splashdust:
This PR reverts gray scale variables in Accordion to the older `--grey-xxx`, since these have working dark mode equivalents. This is a temporary fix until we have proper dark mode tokens in place.


astrit (approved):
Looks good to me ✌️

PR #1354, opened 2024-04-30
Code splitting for Angular and Transitional styles

splashdust:
This PR expands on #1326 by adding even more sub path exports:
* Enable sub-packages in the Angular lib. Each component can now be imported from a sub path.
* Enable per-component registration of transitional styles. Previously, the CSS for all components would get included even if only a few were used. Now it can be selectively included and registered only for the components that are used.


vsjolander (approved):
Awesome to see code splitting getting implemented 🚀 

Looks good to me!

PR #1366, opened 2024-05-07
Dynamic Styles

astrit:
This change is designed to provide a fallback mechanism for applying styles to a web component in environments where Constructed Stylesheets are not supported.

- [x] Grid refactor to use single function to update all variables and process breakpoints 
- [x] Create a controller that injects style using style tag if Constructed Stylesheets are not supported 
- [x] Controller to update when necessary single variables 
- [x] Write tests for Grid
- [x] Use the new style controller on all components


libs/core/src/components/grid/grid.test.ts
```diff
@@ -18,11 +18,39 @@ describe('GdsGrid', () => {
...
   })
 
-  it('sets default properties', () => {
-    const element: GdsGrid = document.createElement('gds-grid') as GdsGrid
-    expect(element.columns).to.be.undefined
-    expect(element.gap).to.be.undefined
-    expect(element.padding).to.be.undefined
-    expect(element.autoColumns).to.be.undefined
+  it('sets columns with single value correctly', async () => {
+    ;(element.columns as any) = '2'
```
vsjolander, line 22: There is something wrong with Prettier. I get these extra `;` that the linting doesn't like... Not sure what causes this.

vsjolander (commented):
Looks good! Great job :)

libs/core/src/components/grid/grid.test.ts
splashdust: Those are required when a line begins with parenthesis. Otherwise it could be interpreted as an invocation of the previous line by the JS parser.

PR #1369, opened 2024-05-10
Fix: dynamic height for full height modals

vsjolander:


splashdust (approved):
Ah, good idea to use `dvh` here!

PR #1371, opened 2024-05-13
`gds-icon` fetch icons from Figma API

astrit:


.changeset/witty-suns-itch.md
```diff
@@ -0,0 +1,5 @@
+---
+"@sebgroup/green-core": minor
+---
+
+Refactor Icon
```
splashdust, line 5: Here it would be good with a bit more information of the new feature, for example "added new icon set from Central" and that icons are used as components

libs/core/src/components/icon/icon.template.ts
splashdust, line 1: I think it would be better to flip the extension of this file to `icon.ts.template`, because otherwise it will get picked up by the compiler and included in dist

libs/core/src/components/icon/icon.ts
```diff
@@ -1,127 +1,26 @@
-import { unsafeCSS } from 'lit'
-import { gdsCustomElement, html } from '../../scoping'
-import { GdsElement } from '../../gds-element'
+import { html, LitElement } from 'lit'
+import { property } from 'lit/decorators.js'
+import { unsafeHTML } from 'lit/directives/unsafe-html.js'
 
-import styles from './stem.styles.scss'
+export abstract class GdsIcon extends LitElement {
+  @property({ type: String }) width = 24
```
splashdust, line 6: Type should be `Number` here, right?
Or the value should be `"24"`

libs/core/webpack.config.js
splashdust, line 1: I guess this is not needed now, since the SVG code is already included in the `.ts`?

splashdust (changes requested):
Overall, superb work!

Found a few minor nit-picks, added as review comments.

And two other things I came to think:
- [ ] Tests
  Probably overkill to generate a test for each icon, but it would be good to pick one of the generated icons and check that the properties (`width`, `height`, `solid`) behave as intended.
- [ ] Accessibility
  Is there a way to set an accessible label for the icon if needed?

libs/core/src/components/icon/icon.template.ts
astrit: This is deleted no need any more!

libs/core/src/components/icon/icon.ts
astrit: Good catch changed the numbers to string ✌️

libs/core/webpack.config.js
astrit: Correct, this is not necessary any more and removed it!

.changeset/witty-suns-itch.md
astrit: Added a recap of changes on the file ✌️

astrit:
@splashdust & @vsjolander the Icon is now ready please check it out and let me know if you have feedback otherwise we could merge this any time now ✌️

splashdust (approved):
🚀 🚀 🚀

PR #1377, opened 2024-05-15
Fonts migration

vsjolander:


astrit (approved):

splashdust (commented):
Looks good to me, but we should remove the changeset before merging.

PR #1379, opened 2024-05-20
Feat/custom dates in calendar

ishmeetseb:


libs/core/src/primitives/calendar/calendar.ts
```diff
@@ -20,6 +20,13 @@ import { watch } from '../../utils/decorators/watch'
 
 import style from './calendar.styles'
 
+type CustomizedDate = {
```
splashdust, line 23: Since this is part of the components API, it should also be exported

splashdust (changes requested):
@ishmeetseb @vsjolander 
Noticed a slight misunderstanding in the way customized dates are configured. My bad for being unclear!

Essentially, it boils down to supporting all customization options on a *per-date* basis. 

For example, the 6th of March could have the red indicator dot, but the 10th of March might need to have the indicator but also be disabled. And I can imagine other similar scenarios coming up in the future, where different dates needs different kinds of customization.

I know it's not specifically needed for the use-case right now, but I think it would be good to get it right from the start.

I've added inline comments on the parts I think needs to change to reflect this. Let me know if it is too unclear and we can go over it on teams instead!

Great work so far, and thanks for contributing!

splashdust (approved):

PR #1391, opened 2024-05-24
Fix regression in Grid auto columns

splashdust:
A recent update broke the feature for setting multiple breakpoints for auto-columns.


astrit (approved):
Looks good to me ✌️

PR #1396, opened 2024-05-27
feat: angular modal animations

vsjolander:
Adding enter and leave animations for modals

No support for reduce motion. Since this is prone for a rewrite to WC I wont dive into that now.


splashdust (approved):
Great stuff! 🥇

Issue #1405, opened 2024-05-30
Icons should have gds-element attribute


vsjolander:
Icons does not have the gds-element attribute which makes it hard to query for the element when scoping is applies.

Current element in DOM:
`<gds-icon-archive></gds-icon-archive>`

Element with attribute:
`<gds-icon-archive gds-element="gds-icon-archive"></gds-icon-archive>`

Not sure if it should have some general attribute for icons aswell making it easy do query for all icons on page? 🤔

splashdust:
I think the `gds-element` should be enough, as you can use attribute prefix selection to do a general query for icon: `[gds-element^="gds-icon"]`
And when dealing with DOM you can just check if its a sub-class of `GdsIcon`

PR #1406, opened 2024-05-30
Add GdsElement back to Icon 

astrit:
As reported #1405


splashdust (approved):
Awesome! 😎

astrit:
Added also the icon preview here 
CleanShot 2024-05-30 at 14 58 55@2x

PR #1413, opened 2024-05-31
fix(tokens): 2016 base darkmode colours

vsjolander:


splashdust (approved):
This looks right to me!

PR #1451, opened 2024-06-19
Dropdown: Add opt-out of mobile view

splashdust:
This PR adds a property to disable to mobile styles for dropdown, since these cause problems in some circumstances (like iframes)


astrit (approved):

vsjolander (approved):
Looks good to me!

PR #1458, opened 2024-06-26
V-Angular

vsjolander:
This is a PR where the Core components of V-Angular, an angular component library based on SEB Vanilla will be migrated to Green.


splashdust:
@vsjolander Looks like we have an issue with importing Green twice for both `.use-green` scope and without. Not sure if it has always been like this or if it is a new check against circular dependencies introduced in newer versions of SASS. Let's investigate further next week!

vsjolander:
> @vsjolander Looks like we have an issue with importing Green twice for both `.use-green` scope and without. Not sure if it has always been like this or if it is a new check against circular dependencies introduced in newer versions of SASS. Let's investigate further next week!

I added this to test green in isolation so i'll just remove it. Having a look at it now.

splashdust (approved):
🚀 🚀 🚀

PR #1461, opened 2024-06-27
Calendar: Promote to public component

splashdust:
The calendar component has existed for a while as an internal primitive used in the Datepicker component. This PR promotes it to a public component available for standalone use.


vsjolander (approved):
Looks good to me 😎👍

PR #1472, opened 2024-07-03
fix(angular): 🐛 not updated dropdown translations on texts input change fix

nerjok:
Somehow with changes previously fixed issue appeared again 

https://github.com/seb-oss/green/pull/705

added fix for current angular library version


splashdust (approved):
Good catch, thank you for the fix!

PR #1479, opened 2024-07-04
Feat/react-segmented-control-component

terrance456:
- React wrapper for green core segmented control 

image


ethan-qsh (approved):

kaylesj4 (approved):

libs/react/src/lib/segmented-control/segmented-control.tsx
```diff
@@ -0,0 +1,65 @@
...
+  disabled?: boolean
+  children?: ReactNode | ReactNode[]
+}
+
+const SegmentedControl: React.FC<SegmentedControlProps> = ({
+  onChange,
+  className,
+  ...props
+}: SegmentedControlProps) => {
+  const onControlChange = (event: Event) => {
```
splashdust, line 45: For consistency, I think `onChange` should just be forwarded directly here, and let the consumer deal with the event. That's how it works in the other wrappers, unless backwards-compatiblity reasons prevent it, and it's also more in line with how events work in native components, like `<input>` and `<button>`.

splashdust (changes requested):
Just found a minor nit-pick, but other than that it looks great. Superb work!

splashdust (approved):
Woho! Thank you for contributing to Green! 💚

PR #1483, opened 2024-07-09
fix: 🐛 visible placeholder value then text changes and nothing selected

nerjok:
if dropdown has  no selected values,  when changing text value visibleValue is calculated according to previous texts, and previous placeholder value displayed. 

To reproduce:

not select options ->change text input, -> old select placeholder displayed,  - forgot about this case in previous pr.


splashdust (approved):
Thank you for contributing to Green 💚

PR #1484, opened 2024-07-09
Popover modality and positioning

splashdust:
This PR adds options to `<gds-popover>` for more granular control over modality, backdrop and positioning.


libs/core/src/components/popover/popover.ts
```diff
@@ -109,6 +111,42 @@ export class GdsPopover extends GdsElement {
...
+
+  /**
+   * When this is set to `true`, the `:backdrop` pseudo-element will be visible if the popover is
+   * in modal mode. When not in modal mode (using the `nonmodal` attribute), this can instead be
+   * set to a selector matching a `<gds-backdrop>` element, in wich case the popover will take
+   * control of that backdrop.
+   *
+   * Example:
+   * ```html
+   * <gds-popover backdrop=".my-backdrop">
```
astrit, line 129: This example here is rendering for some reason on the Public API table 

CleanShot 2024-07-09 at 10 48 53@2x

astrit (approved):

libs/core/src/components/popover/popover.ts
splashdust: Yes, saw that lol! Gonna look into how to escape stuff properly for Storybook.

splashdust: Had just missed to close the code fence block :rage4:

PR #1487, opened 2024-07-10
feat: added a support for accordion so that user can set accordions starting state as open or closed based on their needs.

GauriKudtarkar:
Added a support so that user can set accordions starting state as open or closed based on their needs.

### DIFF
--- a/libs/react/src/lib/accordion/accordion-item.tsx
+++ b/libs/react/src/lib/accordion/accordion-item.tsx
@@ -19,13 +19,13 @@ export interface AccordionItemInterface {
   onClick?: (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
   onOpen?: (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
   onClose?: (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
-  isOpen?: boolean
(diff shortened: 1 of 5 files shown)


libs/react/src/lib/accordion/accordion-item.tsx
```diff
@@ -19,12 +19,13 @@ export interface AccordionItemInterface {
   onClick?: (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
   onOpen?: (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
   onClose?: (e: React.MouseEvent<HTMLButtonElement, MouseEvent>) => void
+  isOpen?: boolean
```
splashdust, line 22: I think this should be named `defaultOpen` to reflec the fact that it can only set the initial state.

splashdust (changes requested):
Looks great, just a minor naming nit pick I found (see comment)

libs/react/src/lib/accordion/accordion-item.tsx
GauriKudtarkar: updated with defaultOpen variable name

splashdust (approved):
Excellent work! Thank you for contributing to Green 💚

PR #1495, opened 2024-07-17
Feature/step list

GauriKudtarkar:
Creating new component named step-list

### DIFF
--- a/libs/core/src/components/step-list/step-list-item.ts
+++ b/libs/core/src/components/step-list/step-list-item.ts
@@ -83,7 +83,7 @@ export class GdsStepListItem extends LitElement {
   @property({ type: String }) title = ""; // Default title is empty
   @property({ type: String }) headerInfo = "";
 
-  updated(changedProperties: PropertyValues<this>) {
+  updated(changedProperties:PropertyValues<this>) {
     if (changedProperties.has("lineColor")) {
       const colorValue = LineColor[this.lineColor as keyof typeof LineColor];
       this.style.setProperty("--line-color", colorValue);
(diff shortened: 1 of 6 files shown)


libs/core/src/components/step-list/step-list-item.ts
```diff
@@ -0,0 +1,133 @@
+import { LitElement, PropertyValues, css } from "lit";
+import { customElement, property } from "lit/decorators.js";
+import "../icon/icons/checkmark.js";
+import { html } from "../../scoping.js";
+
+enum LineColor {
```
vsjolander, line 6: Colors should come from chlorophyll or green-tokens.

This is 2016 design (the current SEB design) so the component should use chlorophyll. Here's [an example how to use the colors from chlorophyll](https://github.com/seb-oss/green/blob/main/libs/core/src/components/button/button.trans.styles.scss).

Probably a good idea to break out the styling to a SCSS file so we keep things seperate for readability.

vsjolander (changes requested):
At a glance it looks good!

We need a couple of things to merge this PR:
- Get colors from Chlorophyll
- Storybook example with some documentation
- Tests to prevent regression

I'm back at work now so we could have a pair programming session if needed. Maybe some of the concepts are new for you so we can figure it out together :)

Great work and thx for contributing! ❤️

libs/core/src/components/step-list/step-list-item.ts
GauriKudtarkar: Hi, I am back from Vacation and sick leave :D hopefully in coming sprint, I could take out some time and fix this issue and then we can also book some pair programming session as you suggested before!

PR #1499, opened 2024-07-19
added react wrappers for green core icons arrowRight, arrowUpRightSqu…

GauriKudtarkar:
…are, arrowWallDown


libs/react/src/lib/icons/arrowUpRightSquare.tsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react'
+import { createComponent } from '@lit/react'
+import { IconSquareArrowTopRight } from '@sebgroup/green-core/components/icon/icons/square-arrow-top-right'
+import { getScopedTagName } from '@sebgroup/green-core/scoping'
+
+export const ArrowUpRightSquare = createComponent({
```
vsjolander, line 6: Name should be `SquareArrowTopRight` to match with icon lib naming.

vsjolander (changes requested):
One small thing. Otherwise great work!

libs/react/src/lib/icons/arrowUpRightSquare.tsx
GauriKudtarkar: Updated file

astrit:
@GauriKudtarkar thank you for the update, will add some other things along the icons and merge it today ✌️

astrit:
@GauriKudtarkar will close this PR in favor of #1522 where your contribution is also included.

PR #1508, opened 2024-08-13
Added props for invalid and required

vsjolander:


astrit (approved):

libs/react/src/lib/datepicker/datepicker.stories.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+export const Invalid: Story = {
+  args: {
+    label: 'Choose date',
+    invalid: true,
+    value: new Date(),
+    testId: 'test',
+  },
+  render: (args) => {
+    console.log(args)
```
believer, line 41: Remove `console.log`?

libs/react/src/lib/datepicker/datepicker.tsx
```diff
@@ -100,6 +121,8 @@ export const Datepicker = ({
     }
   }
 
+  console.log(props)
```
believer, line 124: Remove `console.log`?

believer: @vsjolander There were two different logs, one was not removed https://github.com/seb-oss/green/blob/main/libs/react/src/lib/datepicker/datepicker.tsx#L124

PR #1523, opened 2024-08-26
iOS Tokens for Swift Package Manager

seb-p-dahlback:
Updates the token output for iOS to fulfill the following requirements:

- Correctly formatted Swift
- Package definition and folder structure to conform to Swift Package Manager package structure
- Colors made available via UIKit and SwiftUI in a manner that automatically chooses the dark/light mode color depending on the current environment
- Tree format for token names to make token access a bit cleaner when called from Swift code

This outputs a package that, when copied over to a separate repository, makes it possible for any iOS Swift project to add and use tokens as a dependency.


splashdust (approved):
Looks good to me! Thank you for contributing to Green! 💚

PR #1527, opened 2024-08-26
Dialogs: correct a11y label and modality state

splashdust:


vsjolander (approved):
LGTM!

vsjolander:
@splashdust looks like crypto is not imported and added as a dep. We have [this in extract](https://github.com/seb-oss/green/blob/main/libs/extract/src/lib/id.ts) but maybe better to have a proper lib for these things and deprecate extract when the time comes.

splashdust:
> @splashdust looks like crypto is not imported and added as a dep. We have [this in extract](https://github.com/seb-oss/green/blob/main/libs/extract/src/lib/id.ts) but maybe better to have a proper lib for these things and deprecate extract when the time comes.

Ah, yes, but `crypo` is a browser API, so I'd rather use that than depend on Extract.

Issue #1223, opened 2024-02-21
TextArea info missing bold font


danielgomez-seb:
### Bug already reported?

- [X] I confirm that I have checked if the bug already has been reported

### For which framework/library you are reporting the bug

React

### Component name

TextArea

### Description

The TextArea component is missing bold font on the error message. You can see it in your stories 

https://sebgroup.github.io/green/latest/react/?path=/docs/components-form-textarea--text-area

<img width="776" alt="image" src="https://github.com/sebgroup/green/assets/111365245/68d85aab-43d6-410b-b69e-5fc8d6ce763d">

### Expected Behaviour

Expecting error message to be bolder.

EldRoos:
Found the error, it makes only the last item bold - this time it's the character counter. Will release a fix soon.

EldRoos:
#1529 fixes the bug. Closing!

PR #1529, opened 2024-08-26
Fixed footer always have medium text weight in text area

EldRoos:
#1223


splashdust (approved):
Looking good! 🚀 

Need to add changeset though!

PR #1534, opened 2024-08-27
Update coachmarks

splashdust:
* Add `computeVisibility` callback for customizing visibility
* Remove hard coded dependence on `main` element
* Rename close event to `gds-ui-state` to align with other components

And some other minor tweaks/alignments


astrit (approved):
Looks good to me ✌️

PR #1536, opened 2024-08-28
`green-react` add .js extension to every icon

astrit:
Updated all icons when getting imported from core package with ,js extension and also the build script.


splashdust (approved):
🚀 🚀 🚀

PR #1539, opened 2024-08-28
New component: Floating Action Button (FAB)

splashdust:
The Floating Action Button is used for functions central to the user experience, such as support and help features that needs to be accessible at all times.
 
The button is always floating visible in the users viewport, typically in the lower right corner. This button can be used with an icon only.


apps/core-lib-dev/src/app/datepicker.element.ts
```diff
@@ -3,8 +3,6 @@ import { customElement, state } from 'lit/decorators.js'
 import { html } from '@sebgroup/green-core/scoping'
 
 import '@sebgroup/green-core/components/datepicker/index.js'
-import * as DatepickerTransitionalStyles from '@sebgroup/green-core/components/datepicker/datepicker.trans.styles.js'
```
vsjolander, line 6: Was this intentional?

apps/core-lib-dev/src/app/calendar.element.ts
```diff
@@ -3,8 +3,6 @@ import { customElement, state } from 'lit/decorators.js'
 import { html } from '@sebgroup/green-core/scoping'
 
 import '@sebgroup/green-core/components/calendar/index.js'
-import * as CalendarTransitionalStyles from '@sebgroup/green-core/components/calendar/calendar.trans.styles.js'
```
vsjolander, line 6: Was this intentional?

vsjolander (commented):
LGTM!

not sure if the removal of transitional styles on other components were intentional?

apps/core-lib-dev/src/app/calendar.element.ts
splashdust: Yes, this is just in `core-lib-dev` in order to test without transitional styles!

PR #1546, opened 2024-08-30
feat: react tooltips, tooltips for table

samspi:
* Added tooltip link component
* Added tooltip button component
* Implemented tooltip feature in Table


splashdust:
@samspi Thank you for the PR!

I added a story for the tooltip and updated the docs for Table. Also, I removed the `TooltipButton` component (or rather moved it locally to table, where it is used internally), as I feel that it was a bit too specific to export as a public part of the library.

libs/react/src/lib/table/story/multiple-story-tables.tsx
```diff
@@ -124,7 +124,7 @@ export const SortableTable = () => {
       <TableHeader>
         <TableRow>
           {columnData.map((column: SuperHeroColumnDataT, i: number) => (
-            <TableHeaderCell key={i} {...column} />
+            <TableHeaderCell key={i} {...column} tooltipText={'Tooltip'} />
```
vsjolander, line 127: Its better to pass children to `TableHeaderCell`  here:
```jsx
<TableHeaderCell key={i} {...column}>
    <TooltipButton tooltip={tooltipText} className="sg-table-sort">
        {children}
    </TooltipButton>
</TableHeaderCell>
```
and remove the logic from the Table component. That way we keep components seperate and minimize dependencies.

vsjolander (changes requested):
Hi and thank you for your contribution!

We (The Green Design System team) have found some issues that needs to be addressed:

## Dynamic width
Long messages get very high:
Screenshot 2024-09-03 at 11 00 04

The width should be dynamic (follow the width of the text) but still follow the [writing guidelines](https://designlibrary.sebgroup.com/components/component-tooltip).

## Accessibility issues
- Tooltip only trigger on mouse hover - needs to support keyboard navigation
- Tooltip is placed outside of screen, see [storybook](https://storybook.seb.io/pr-1546/react/?path=/story/components-tooltip--tooltip) we are using floating UI to handle this.

libs/react/src/lib/table/story/multiple-story-tables.tsx
samspi: Makes sense! I'll get started on that!

vsjolander: How is this going @samspi? Can we assist you in any way?

samspi: Hi there. An update. I've been unable to dedicate time to these fixes and I'm looking into handing it over to another member in the team. Thanks for your patience!

samspi:
Hi there. Another much belated update. Again my apologies for the lack of responses. I will be closing pull request for now and reopening it when I have time to dedicate myself to these fixes.

PR #1551, opened 2024-09-02
Feature/fix double scroll issue for modal

GauriKudtarkar:


splashdust:
@GauriKudtarkar This PR is reverting lots of old changes and have lots of conflicts. This can happen if you have for example rebased and done a force push. Then you may end up with rewritten history that doesn't match the remote.

The easiest way to solve is to close this PR and open a new from the latest commit on the Green main branch.

PR #1553, opened 2024-09-03
updated modal related files to make sure the double scroll bar will n…

GauriKudtarkar:
fix: modal double scroll bar issue fixed

### DIFF
--- a/libs/chlorophyll/scss/components/modal/_index.scss
+++ b/libs/chlorophyll/scss/components/modal/_index.scss
@@ -2,6 +2,11 @@
 @use 'mixins';
 
 // base styles for modal
+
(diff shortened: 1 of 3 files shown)


splashdust (changes requested):
Hey @GauriKudtarkar! In the Angular version of modal, `body-scroll-lock` is used to achieve this effect. This package also handles platform variations and makes sure it works on all devices and without resetting scroll position (as `overflow: hidden` will do). So, for consistency, I think we should use the same solution here.

Here you can find instructions on how to use it: https://github.com/willmcpo/body-scroll-lock

splashdust:
Continued in #1572

PR #1554, opened 2024-09-03
Menu button: Expose CSSPart for main slot

splashdust:
There was a request to be able to customize text-alignment of the menu button. Usually, this can be done by styling the slotted element, but in the menu button the slot itself needs some default styles which sets it to `display: inline-block`.

This PR adds a CSSPart hook for the main slot so that it can be customized.

### DIFF
--- a/.changeset/four-beds-accept.md
+++ b/.changeset/four-beds-accept.md
@@ -1,0 +1,5 @@
+---
+"@sebgroup/green-core": minor
+---
+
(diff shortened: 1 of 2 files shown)


vsjolander (approved):
LGTM!

I think this should be a prop/attribute on the button aswell like `justify-content="center/space-between"`.

But that will also control the positioning of lead and trail slots.

But still having this makes sense.

astrit (approved):

PR #1563, opened 2024-09-04
Rename all links from `sebgroup/green` to `seb-oss/green`

astrit:
Rename all links from `https://github.com/sebgroup/green` to `https://github.com/seb-oss/green`  

> This change doesn't affect the package name that still is `@sebgroup/green...`
> Only the direct Github links have been renamed


splashdust (approved):
Awesome!

PR #1566, opened 2024-09-06
Fix Improve windows development

fredrik-oberg-seb:
Intention is to make it easier for SEB developers on windows computers to get started with this project.
Since Yarn is not an recommended option at SEB we instead of `yarn` use `npm run` to run node scripts.

Also changed to use the `copyfiles` instead of Linux command `cp` . 
`copyfiles` is an existing dependency in the project and works on both mac/linux and windows computers.   

References #725


splashdust (approved):
Looks good to me! Great initiative to change these, and thank you for contributing to Green! 💚

PR #1567, opened 2024-09-06
moving sass declarations before nested rules

GauriKudtarkar:
solving issue of sass:
image


splashdust (approved):
Good catch, look good to me!

PR #1569, opened 2024-09-06
Feature: An input property to be able to validate the input before going to the next page in the in page wizard

PierreSEBank:
An input property to be able to validate the input before going to the next page.

errormsg


splashdust (approved):
Looks good to me! Great work, and thank you contributing to Green! 💚

PR #1572, opened 2024-09-09
hide background scrollbar using body-scroll-lock

GauriKudtarkar:
fix for hiding background scroll


splashdust:
@GauriKudtarkar Looks good to me! It would be good if you also added a changeset though. If the PR is merged without this, no release will be published. Check the Changeset Status comment from the bot for instructions. You can also read here: https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md

PR #1574, opened 2024-09-09
Update: `gds-icon`, `gds-badge` and login demo

astrit:
This PR updates `icons` and some other small tweaks: 

- [x] New icons added on `core`
- [x] New icons added and generated on `react`
- [x] Core App updated and 
- [x] Renamed and added the brand icons and updated their paths in the repo.
- [x] Storybook also updated to get all icons automatically so no need to update manually
- [x] Group Form component stories 
- [x] Add back `gds-badge` story and update component


astrit:
Badge updates: 

CleanShot 2024-09-10 at 22 21 03@2x

libs/core/src/components/badge/badge.ts
```diff
@@ -0,0 +1,74 @@
+import { html } from 'lit/static-html.js'
+import { unsafeCSS } from 'lit'
+import { gdsCustomElement } from '../../utils/helpers/custom-element-scoping'
+import { GdsElement } from '../../gds-element'
+import { tokens } from '../../tokens.style'
+import { styleExpressionProperty } from '../../utils/decorators/style-expression-property'
+
+import BadgeCSS from './badge.style.css?inline'
```
splashdust, line 8: Here, I would
* rename this file to `badge.style.ts` to align with other components
* remove the `inline` query, since that is for static files
* remove `unsafeCSS` below, since we're using the `css` template tag from Lit

libs/core/src/components/icon/icons/seb.ts
splashdust, line 1: Looks like some icons were renamed. When this happens, we need to keep the old name and mark it as deprecated, otherwise it will be a breaking change. I know several people are using the SEB icon already, so removing it would cause some frustration.

splashdust (commented):
Amazing work as usual!

Just two things:
* Minor note on how the badge CSS is imported. Not super important.
* Icon deprecation. Although it would have relatively minor impact to rename icons this early on, we still need to solve this now I think. When an icon is renamed, the old one needs to remain and be marked as deprecated. Maybe this needs to be added to the script, or it could be a manual process.

I think the only immediate blocker on this PR is the SEB icon though, the rest we could solve in a new PR.

libs/core/src/components/icon/icons/seb.ts
astrit: Added back the icon and marked it as `deprecated`

splashdust (approved):
🚀🚀🚀

PR #1587, opened 2024-09-12
Fix green jest config with multiple transform ignore patterns

### LINKED ISSUE #1586: Regex in /libs/core/green-jest-config.cjs transformIgnorePatterns
### Bug already reported?

Core

### Component name

_No response_

### Description

When trying to use the new green-jest-config for Jest we are getting error `SyntaxError: Unexpected token 'export'` 

After some digging around the Jest docs and code I see that when providing multiple `transformIgnorePatterns` regex then each regex will need to match otherwise it will transform the file. 

Example: 
We have this file: `node_modules\\@common-shell\\shell-site-api\\dist\\index.js` that should be ignored from transformation.

In our `jest.config.transformIgnorePatterns` we have the regex to make our Jest test work:
```regex
/node_modules\\(?!.*\.mjs$|.*\.esm$|lit-html|lit|@lit|@sebgroup\\green-core|@sebgroup\\extract|@common-shell\\shell-site-api|flat\\)/` 
```

So when we Jest run that one against the file it returns false 
```javascript
> /node_modules\\(?!.*\.mjs$|.*\.esm$|lit-html|lit|@lit|@sebgroup\\green-core|@sebgroup\\extract|@common-shell\\shell-site-api|flat\\)/.test('node_modules\\@common-shell\\shell-site-api\\dist\\index.js')
false
```

But when merging the regex from `green-jest-config.cjs` we get this regex
```regex
/node_modules\\(?!.*.mjs$|@sebgroup\\green-react|@sebgroup\\extract|@sebgroup\\green-core|lit-element|lit-html|lit|@lit|@lit-labs\\)|node_modules\\(?!.*\.mjs$|.*\.esm$|lit-html|lit|@lit|@sebgroup\\green-core|@sebgroup\\extract|@common-shell\\shell-site-api|flat\\)/`
```
And running that one will return true. 

```javascript
> /node_modules\\(?!.*.mjs$|@sebgroup\\green-react|@sebgroup\\extract|@sebgroup\\green-core|lit-element|lit-html|lit|@lit|@lit-labs\\)|node_modules\\(?!.*\.mjs$|.*\.esm$|lit-html|lit|@lit|@sebgroup\\green-core|@sebgroup\\extract|@common-shell\\shell-site-api|flat\\)/.test('node_modules\\@common-shell\\shell-site-api\\dist\\index.js')
true
```
Jest is merging the `transformIgnorePatterns` with an `|` "OR" and then all expression will need to match.
 

 

### Steps To Reproduce

_No response_

### Current Behaviour

_No response_

### Expected Behaviour

_No response_

fredrik-oberg-seb:
Fixes #1586 

Join patterns with `&` "AND". 
This is needed so the project that is using green-jest-config.cjs can have there own transform ignore patterns. 

Jest is joining the patterns with `|` "OR" and if there is multiple then all need to ignore the same patterns.


splashdust (approved):
Good catch, and yeah this makes sense! I'm a little confused as to why Jest would join the expressions with an OR operator by default though. I wonder if there is a reason for it, or if it's just an oversight.

In any case, this change appears to make sense here, so let's merge it. I guess we'll find out if it causes some other issue one way or another :)

PR #1602, opened 2024-09-18
Add support for pseudo-selectors in Style Expression Properties

splashdust:
This PR adds support for specifying pseudo selectors like `:hover`, `:focus`, etc, in Style Expression Properties.

For example: 
```html
<gds-card padding="xs; hover: xs s; >l { s m; hover: xl }"></gds-card>
```

Sub-selectors are specified by prefixing a value sequence with `xxx:`, and selector/value sequences are terminated by `;`
The sub-selector will then be appended/merged with the base selector, ie `:host(:xxx)` or `.some-sel:xxx`. The `;` is only mandatory when there are more than one sequence, meaning that this upgrade will be backwards-compatible with existing expressions.

Here is a slightly more realistic example, changing the border and padding on hover:

```html
<gds-card
  radius="s"
  border="4xs/l2-stroke-primary; hover: s/l2-stroke-primary"
  padding="xl; hover: l"
  background="l2-background-secondary"
>
...
</gds-card>
```

### DIFF
--- a/.changeset/wicked-rules-camp.md
+++ b/.changeset/wicked-rules-camp.md
@@ -1,0 +1,5 @@
+---
+'@sebgroup/green-core': minor
+---
+
+**Declarative Layout:** Add support for pseudo sub-selectors, like `:hover`, ':focus', etc.
--- a/libs/core/src/storybook-docs/concepts/declarative-layout.mdx
+++ b/libs/core/src/storybook-docs/concepts/declarative-layout.mdx
@@ -82,8 +82,17 @@ a CSS variable. For example: `padding="xs; l {m xl m xl}"`
 When no media query is specified, it is the same as setting the breakpoint to 0. So, `padding="0 {m}"` is equivalent
 to `padding="m"`.
 
-Semicolon `;` is used to separate media queries, but is only mandatory when the shorthand 0 breakpoint is used, as
-in the first example above.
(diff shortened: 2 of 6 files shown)


vsjolander (approved):
LGTM!
I have a feeling that we are making this too complicated and it will be tech debt... But it's at the same time soo good and creates an awesome DevEx! ☯️

splashdust:
> I have a feeling that we are making this too complicated and it will be tech debt... But it's at the same time soo good and creates an awesome DevEx! ☯️

Yes, it's a fine line! I think that we currently have a pretty good balance between complexity and value added, and we need to keep it that way. Let's discuss this further in light of the current feature set, and maybe we can refine a more final set of requirements, now that we have something concrete to look at.

As for this PR, most of the changes are just refactoring the original implementation to make it more maintainable and easier to understand. The new bit is pretty much just this: https://github.com/seb-oss/green/pull/1602/files#diff-d1a884735b6a1407454072cd84d9ba12577a8311d6a87f5be1b0a4f60dea2adcR107-R111

And these changes to how the final CSS is rendered: https://github.com/seb-oss/green/pull/1602/files#diff-d1a884735b6a1407454072cd84d9ba12577a8311d6a87f5be1b0a4f60dea2adcR107-R111

PR #1606, opened 2024-09-24
chore: build script for core react wrappers & build core wrappers

vsjolander:
This PR adds:
- a build script for React wrappers of the core components 
- NX target using the build script
- Built react wrappers


libs/core/src/utils/decorators/style-expression-property.ts
```diff
@@ -44,7 +44,7 @@ export function styleExpressionProperty(
     const sel = options?.selector ?? String(':host')
     const prop = options?.property ?? String(descriptor)
     const valueTemplate =
-      options?.valueTemplate ?? ((v) => `var(--gds-space-${v}, 0)`)
```
splashdust, line 47: Removing the fallback again, is this intentional?

.changeset/proud-olives-repeat.md
splashdust, line 1: Shouldn't there be a changeset for React here?

PR #1607, opened 2024-09-24
gds-text to use internal classes instead of :not() selectors

vsjolander:


splashdust (approved):
Looks good to me!

At some point down the line, we could instead put declarative styles on a separate cascade layer, and any default styles on a default layer, and then have the declarative layer override the default layer. That way, no extra processing would be required in the templates. But for now, using classes works!

PR #1608, opened 2024-09-24
Style expression fallback & story for container padding

vsjolander:


libs/core/src/components/layout/container/container.ts
```diff
@@ -114,15 +111,19 @@ export class GdsContainer extends GdsElement {
    * The above example will apply the padding style of `xs` for `small` devices, `m` for `medium` devices, and `xl` for large devices.
    *
    */
-  @styleExpressionProperty()
+  @styleExpressionProperty({
+    valueTemplate: v => {
+      return `var(--gds-space-${v}, 0)`
+    }
```
splashdust, line 117: I don't think this is needed, as the default valueTemplate already interpolates to space tokens: https://github.com/seb-oss/green/blob/main/libs/core/src/utils/decorators/style-expression-property.ts#L46

It doesn't have a fallback though, so that might be good to add to the default template. An alternative to `0` could be to just add the raw value as fallback, so that any custom value would be accepted; `var(--gds-space-${v}, ${v})`

vsjolander: ah! didn't know that. I prob added the wrong value and then I tought this was needed. adding the fallback instead. Thanks!

splashdust (approved):
👌

PR #1628, opened 2024-10-02
Benchmarking + some perf. improvements

splashdust:
This PR adds a new app to the monorepo, `core-lib-benchmark`. It is a simple web app and a script to collect and compare Chrome traces using Puppeteer. The app is just a simple web app with a router, where each route can contain a set of components or layout that should be tested.

Benchmarks can be run like this:
```
$ nx run core-lib-benchmark:collect-trace decl-layout-1
```

This will fire up a dev server, and point the headless Chrome instance to the `/decl-layout-1` route. The trace will be outputted as JSON to the `./trace-snapshots` dir. The trace can be used as comparison in future runs, or opened and inspected in DevTools.

This PR also adds a caching optimization for constructed stylesheets, and replaces the use of the `watch` decorator with a setter in the `@styleExpressionProperty` decorator.

Example output, comparing difference between using cached stylesheets, versus no caching:
```
Scripting: 843.55ms
Rendering: 504.05ms
Painting: 4.93ms
-------------------
Total: 1352.53ms

Compared to snapshot-decl-layout-1-m3pro:

Scripting: 82.70%
Rendering: 577.22%
Painting: 10.27%
-------------------
Overall: 150.19%
```

Other minor things:
Remove dyanmic line-height from input, since that is a textarea feature
Remove `inline-size` style from badge, since it overrode `width` and caused broken layout.


vsjolander (approved):

vsjolander:
Amazing work like always 👌

PR #1631, opened 2024-10-04
Better react slider examples

EldRoos:
Slider storybook for react now shows label and sublabel, along with default value and a big number example.

Will make it easier to understand and bug test slider in react. Now is much more similar to the examples in Angular.


splashdust (approved):
👌

Issue #1603, opened 2024-09-20
Variable Width on Segmented Control


astrit:
### Description

Make it possible to have specific width on segmented controls

PR #1636, opened 2024-10-07
Refactor Segmented Control

splashdust:
This change refactors a large chunk of the Segmented Control implementation:

- Rewrite layout code to use `css-scroll-snap`
- Remove `seg-min-width`
- Add style expression properties to `gds-segment` for size control

Related to #1603

### DIFF
--- a/libs/core/src/components/segmented-control/segmented-control.style.css
+++ b/libs/core/src/components/segmented-control/segmented-control.style.css
@@ -5,7 +5,6 @@
   box-sizing: border-box;
   contain: layout;
   display: inline-flex;
-  /* gap: 0.25rem; */
   height: 3rem;
   width: 100%;
   position: relative;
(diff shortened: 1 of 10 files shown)


libs/core/src/components/segmented-control/segmented-control.style.css
```diff
@@ -5,9 +5,9 @@
   box-sizing: border-box;
   contain: layout;
   display: inline-flex;
-  gap: 0.25rem;
   height: 3rem;
```
astrit, line 8: I can see here we are using rem's, should we try to switch other components to using rem units or this to PX ?

astrit (approved):
Looking good to me, nice update 👌

libs/core/src/components/segmented-control/segmented-control.style.css
splashdust: Typical best practice is to use rem's for sizing stuff. That way, when a user has changed the base font-size in their browser, other stuff will follow proportionally. This needs more testing though!

PR #1639, opened 2024-10-10
Refactor how `gds-element` is set

splashdust:
This PR refactors how the `gds-element` attribute is set on each gds element. Now it is set by the `gdsCustomElement` instead of on `connectedCallback()`. This makes it work even if the component is registered under a different element name, and also improve performance a bit since it no longer needs a lookup in the registry for each component instance.


vsjolander (approved):
LGTM!

Would like a rundown in person as well 😄

PR #1649, opened 2024-10-14
Refactor the element lookup table to a global singleton

splashdust:
This PR refactors the element lookup table as a global singleton rather than a module-scoped object. This should help in module federated scenarios where code has been split up into multiple dynamically loaded modules.

An added benefint is that it now becomes easier to troubleshoot scoping related issues, since you can simply inspect the singleton:
image


astrit (approved):
Looks good to me 👌

PR #1676, opened 2024-10-16
Storybook port change

EldRoos:
React now have fort 4401 as default and Angular 4402. So they don't have same as chlorophyl on 4400.

### DIFF
--- a/.changeset/unlucky-actors-perform.md
+++ b/.changeset/unlucky-actors-perform.md
@@ -1,0 +1,6 @@
+---
+'@sebgroup/green-angular': patch
+'@sebgroup/green-react': patch
+---
+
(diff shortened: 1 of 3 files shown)


.changeset/unlucky-actors-perform.md
splashdust, line 1: No need for a changeset when the libraries have not changed. This will trigger new releases without any changes.

EldRoos:
Removed changeset and now it fails! @splashdust

splashdust:
> Removed changeset and now it fails! @splashdust

That's OK in this case, since we don't need the changeset. You can also create an empty changeset to get rid of the check error, but it's not mandatory.

Issue #1171, opened 2024-01-24
Accordian: Should be possible to select aria-level


PasiVaisanen:
### Does the component already exist

- [X] I confirm that I have checked if the component already exists in any of the Green libraries

### For which framework/library you are requesting the component

Angular

### Description

Currently the aria-level is hardcoded to 2, it should be possible to select the aria-level depending of the content where the accordian is displayed.

### Who will benefit from this feature?

Hard to use a speech reader when we cannot select the area-level.

PR #1692, opened 2024-10-22
Accordion: Can set aria-level in angular

EldRoos:
The aria-level (heading level) can now be set in Angular. Same as React can.
Closes: #1171


splashdust (approved):
👌

PR #1697, opened 2024-10-23
Form control commonalities

splashdust:
This PR adds reusable primitives for internal form control layout. For now, only the header and footer are replaced in `gds-input` and `gds-textarea`. Next step will be to do the same for the field design, and to implement the primitives in more of the form controls.


libs/core/src/primitives/form-control-footer/form-control-footer.ts
```diff
@@ -0,0 +1,59 @@
...
+} from '../../utils/helpers/custom-element-scoping'
+import { styles } from './form-control-footer.styles'
+
+import '../../components/badge'
+import '../../components/icon/icons/triangle-exclamation'
+
+/**
+ * @element gds-form-control-footer
+ */
+@gdsCustomElement('gds-form-control-footer')
```
astrit, line 17: I would suggest to name this to `gds-field-footer` and same pattern for header, since it is a bit more clear on what it is, also in design perhaps could be a bit more easy to understand.

libs/core/src/primitives/form-control-header/form-control-header.ts
```diff
@@ -0,0 +1,110 @@
...
+import '../../components/card'
+
+/**
+ * @element gds-form-control-header
+ *
+ * @slot label - The label of the form control.
+ * @slot supporting-text - The supporting text of the form control.
+ * @slot extended-supporting-text - The extended supporting text, or "Show more info" card, of the form control.
+ */
+@gdsCustomElement('gds-form-control-header')
```
astrit, line 23: Same here would suggest to name `gds-field-header` ✌️

libs/core/src/primitives/form-control-footer/form-control-footer.ts
splashdust: I would argue that `form control` is the technically correct term that we should stick to in the context of forms.

For reference, W3C defines it as "An HTML form is a section of a document containing normal content, markup, special elements called [controls](https://www.w3.org/TR/html401/interact/forms.html#form-controls)"

PR #1672, opened 2024-10-15
Added screenreader text on close button

EldRoos:
Added screen reader text on close button.
Connected to this issue: #706 & #665

Please check if you also get visual error on AlertRibbon close button on colored backgrounds in React:
https://storybook.seb.io/latest/react/?path=/docs/components-alertribbon--docs

Locally it looks like this line is not triggered anymore:
green\libs\chlorophyll\scss\components\alert-ribbon\_mixins.scss
line 139 - background-color: var(--color);

Also green\libs\react\src\lib\card\card.stories.tsx file shows up as red/error?

PR #1718, opened 2024-10-29
Alert Ribbon Close Button Fix

EldRoos:
Both Chlorophyll and React close button now have custom label and shows up correctly in screen readers.

Check if closeText, isCloseable, closeAriaLabel & closeButton are correct.

Note: This is a remake and a split up of #1672.

### DIFF
--- a/libs/chlorophyll/scss/components/alert-ribbon/alert-ribbon.stories.js
+++ b/libs/chlorophyll/scss/components/alert-ribbon/alert-ribbon.stories.js
@@ -27,6 +27,12 @@ export default {
     closeButton: {
       control: 'boolean',
     },
+    isCloseable: {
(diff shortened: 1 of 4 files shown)


astrit:
@EldRoos 

You had a test failing, The issue is that the `closeAriaLabel` prop was not being used in the AlertRibbon component. Instead, the closeText prop was being used to set the aria-label for the close button. 

I fixed that so check if everything in order and should be good to go! 

Nice job 👌

EldRoos:
Great Astrit. I was thinking this fix when I was in the shower :) but you beat me to it. I wonder why we use 2 variables for same thing.

splashdust (approved):
Nice work! 👌

PR #1725, opened 2024-10-30
Modal react close text fix

EldRoos:
All modals for React can now have custom close button text for screen readers.
(Will do modal fix for Angular next)


EldRoos:
Ok react autotest fails. Will check that out.

splashdust (approved):
👌

PR #1732, opened 2024-11-01
updated transloco version

pontus-github:
`@ngneat/transloco` has been deprecated and will not recieve updates anymore. Transloco lib has been moved to to new scope and is now available under `@jsverse/transloco`. This PR updates the scope used for transloco from `@ngneat` to `@jsverse`, which also means a major version upgrade of transloco from v6 to v7, as v6 is not available under the new `@jsverse` scope.


splashdust:
@pontus-github This PR has an empty changeset, so it will not trigger a new release. Is this intentional?

pontus-github:
> @pontus-github This PR has an empty changeset, so it will not trigger a new release. Is this intentional?

Yes, this is intentional. Since the changes only effect v-angular components, it was agreed with @vsjolander that empty changeset would be appropriate

splashdust (approved):

PR #1738, opened 2024-11-01
Update read-before-using-angular.mdx

vsjolander:


rimantask (commented):
Text looks looks good. Diagram misses alternative path for "Has all the required components" and "use angular green wrapper" should not be a condition figure.

libs/angular/src/v-angular/read-before-using-angular.mdx
```diff
@@ -4,6 +4,57 @@ import { Meta } from '@storybook/addon-docs'
 
 # Read this before using V-Angular
 
-### V-Angular is being migrated to Green Design System and receive some updated but not be actively developed. Migrate to Green Core (or Green Angular and Green React)
+V-Angular is being migrated to Green Design System (GDS), thus if you want to use v-Angular components, you can use [https://github.sebank.se/DesignLibrary/v-angular](https://github.sebank.se/DesignLibrary/v-angular) for now and once the migration is finished migrate your applications to V-Angular which will support GDS. V-Angular could be used in cases when you don't find corresponding components in Green Core.
```
NamitaMalik, line 7: Discussed this disclaimer again with @rimantask, it should be updated to:

> _v-angular_ is being migrated to Green Design System (GDS), thus if you want to use _v-angular_ components, you can use https://github.sebank.se/DesignLibrary/v-angular for now and once the migration is finished, migrate your applications to Green's V-Angular which will support GDS. Green's V-Angular should be used in cases when you don't find corresponding components in Green Core.
It was decided to migrate _v-angular_ to GDS in order to provide a more consistent and unified experience across all SEB products. The migration is still in progress and the components are being updated to match the new design system.

rimantask: Just realized that link to v-angular is internal link, thus we should either state "for SEB employees only" at the beginning, or replace link by simply "them" meaning that "you can use v-angular components for now". But anyway mentioning v-angular is meaningless to external users.

PR #1741, opened 2024-11-04
Implement selective transitional styles

splashdust:
This PR adds a property, `designVersion`, to `gds-theme`, which can be used to selectively control the design version (2016 or 2023) for its descendants. This should also fix the broken design version switch in Storybook.


vsjolander:
Lookin guuud! Have two thoughts:
- Should it rather be `token-version` instead of `design-version`?
- Should it be possible to manually register components as `2016` or `2023` so you can mix and match?

splashdust:
> Lookin guuud! Have two thoughts:
> 
> * Should it rather be `token-version` instead of `design-version`?

I think `design-version` makes sense, since this affects which stylesheets are used in components, and not just tokens.

> * Should it be possible to manually register components as `2016` or `2023` so you can mix and match?

Yes, this update makes this possible! The transitional styles can be loaded individually, or in bulk, just as before, but `gds-theme` can now be used to control where they get applied. So if you need to turn it on/off for an individual components, just wrap in a `gds-theme` and set the `designVersion` prop. I can demo later how this works.

vsjolander (approved):

PR #1747, opened 2024-11-05
Angular modals custom close button label

EldRoos:
Inc slideout and takeover.


libs/angular/src/lib/modal/modal.component.ts
```diff
@@ -91,6 +91,7 @@ import { NggModalHeaderDirective } from './modal-header.directive'
 export class NggModalComponent implements OnDestroy, OnInit {
   @Input() public modalType?: ModalType
   @Input() public header?: string
+  @Input() public closeText: string = "Close Modal"
```
splashdust, line 94: @EldRoos Linting is complaining about this one due to the type being specified even though it is easily inferred (since it is set to a string value). You can just remove the `:string` part

splashdust (changes requested):

splashdust (approved):
👌

PR #1765, opened 2024-11-07
Upgrade to nx 20 and Angular 17

splashdust:
This PR upgrades the base Angular dependency to version 17. This is considered a breaking change, since consumers will need to update their peer dependencies to match if they are still below 17.


vsjolander (approved):
Great work! Looks good to me

PR #1804, opened 2024-11-18
add an event dispatch when search has been cleared

Ateezey:
Change so that the consumer of the component can listen to the clear button click


splashdust (changes requested):
Looks good, just have two small things:
- Name the event `gds-input-cleared` to signify that it is a custom event from Green
- Add to doc header for the component, like `@event gds-input-cleared - explanaition`

splashdust (approved):
👌

PR #1814, opened 2024-11-20
Fix FAB `max-width`

astrit:


splashdust (approved):
Looks super good! Would be good to add a changeset for the FAB fix also.

libs/core/src/components/container/container.ts
```diff
@@ -536,6 +536,17 @@ export class GdsContainer extends GdsElement {
...
+  /**
+   * Controls the `box-sizing` property.
+   * Supports only token
+   *
+   * @property box-sizing
+   */
+  @styleExpressionProperty({
+    valueTemplate: (v) => v,
+  })
+  'box-sizing'?: string
```
vsjolander, line 548: We should generate types from our tokens so we can use them here!

vsjolander (approved):

PR #1817, opened 2024-11-21
feat: update v-angular

dev3072:


libs/angular/src/v-angular/checkbox/checkbox.stories.ts_
```diff
@@ -33,7 +34,7 @@ export default {
       providers: [importProvidersFrom(NgvI18nModule)], // Changes: removed
     }),
     moduleMetadata({
-      imports: [CommonModule, FormsModule, ReactiveFormsModule, NgvI18nModule],
+      imports: [CommonModule, FormsModule, ReactiveFormsModule, NgvI18nModule, FontAwesomeModule],
```
maoq01, line 37: Reference and import of `FontAwesomeModule` can be removed.

PR #1820, opened 2024-11-21
Feat FAB add `z-index` property

astrit:


splashdust:
Here's what we discussed related to this PR:
It would be good to think of and discuss a broader general strategy for z-stacking in the design system. Until we have done that, it may be premature to add this property, so we decided to just hard-code a high-ish value for now and not expose it though public API. That way we can change it later without breaking API changes.

astrit:
> Here's what we discussed related to this PR: It would be good to think of and discuss a broader general strategy for z-stacking in the design system. Until we have done that, it may be premature to add this property, so we decided to just hard-code a high-ish value for now and not expose it through public API. That way we can change it later without breaking API changes.

I did remove the `z-index` prop and updated the example to better represent what is happening and a explanation that the `z-index` is hard coded to `1050` 

Also I did pass the styling to the host it self for positioning instead since if fab is used within a flex then the host takes the space when using gap. 

@splashdust ping!

vsjolander (approved):
LGTM!

PR #1833, opened 2024-11-22
Add concept documentation about tokens

splashdust:
This PR adds a Tokens section under Concepts, and also fleshes out the Colors documentation a bit more.

It also includes som fixes to GdsTheme:
- Fix a bug where light-mode variables remained when switching to dark mode
- Add the rest of the tokens. Previously `GdsTheme` only included color tokens. Check the new Tokens doc to see why this is useful.


vsjolander (approved):
Great work!! 💎 🚀  LGTM 👍

Issue #1442, opened 2024-06-14
Datepicker: year loops around


arekon7:
### Bug already reported?

- [X] I confirm that I have checked if the bug already has been reported

### For which framework/library you are reporting the bug

Angular

### Component name

Datepicker

### Description

Datepicker calendar "year" loops around. When navigating with "next" button till next year, it shows "1924" as a next year. 

Please check video:
[Dp-year-issue.webm](https://github.com/seb-oss/green/assets/40074601/fdbb74fd-720c-416c-99f2-505cfd6665f2)

### Steps To Reproduce

Steps to reproduce were written on 2024-06-14:

1. Go to https://storybook.seb.io/latest/angular/?path=/docs/components-datepicker--documentation#datepickeroptions page
2. Open datepicker "Birthday"
3. Using "Next" arrow navigate through months till the next year

### Current Behaviour

Next year is shown as 1924 (first option form the years list)

### Expected Behaviour

Disabled over looping to the first year option from list.

EldRoos:
I can confirm. It's same on the "normal" datepicker **Date**. If you go over 2034 you go to 2014.

PR #1852, opened 2024-11-28
Datepicker: Increasing max year will not overflow to min year

EldRoos:
Fixes: #1442


splashdust (changes requested):
The change itself looks good to me, but a test to guard against regressions would be good!

I.e, create a test which sets the date to the last month of the max year and then clicks the next button and verify that year did not change, and then same for min (or vice versa). Could also be split in two tests.

libs/core/src/components/datepicker/datepicker.test.ts
```diff
@@ -513,5 +514,34 @@ describe('<gds-datepicker>', () => {
         onlyDate(new Date('2024-01-01')),
       )
     })
+
+    it('should not overflow the year when trying to increase max year', async () => {
```
splashdust, line 518: It would be good to test the previous month action too. That could be done in a separate test

libs/core/src/components/datepicker/datepicker.test.ts
```diff
@@ -513,5 +514,34 @@ describe('<gds-datepicker>', () => {
...
+        keyPress = 'Shift+Alt+Tab'
+      }
+      
+      await sendKeys({
+        press: keyPress,
+      })
+    
+      await sendKeys({
+      press: 'Enter',
+      })
```
splashdust, line 539: Here, I think it would be easier to just target the next month button directly.

Something like this:
```ts
const nextMonthButton = el.shadowRoot!.querySelector<GdsButton>(
  `${getScopedTagName('gds-button')}[aria-label="Next month"]`,
)!

nextMonthButton.click()
```

That will also make it easier to write a test that checks the previous month action.

splashdust (changes requested):

libs/core/src/components/datepicker/datepicker.test.ts
EldRoos: Fixed

libs/core/src/components/datepicker/datepicker.test.ts
EldRoos: Fixed

splashdust (approved):
Nice work! 💯

PR #1854, opened 2024-12-02
Tokens: Remove Color V1 and update all components using that version

astrit:
Intent behind this change is to eliminate the use/import of ref variables and the V1 color tokens. 
In turn we have a all components aligned and also hopefully eliminate some bytes by not importing all tokens. 

This PR updates the following:

1. Tokens 
2. Theme import tokens for light and dark mode
3. Calendar
4. Datepicker
5. Dropdown 
6. Summary
7. Segment
8. Option
9. Menu Heading

TODO: 
Remove tokens also from tokens package


libs/core/src/components/datepicker/datepicker.trans.styles.scss
```diff
@@ -88,7 +88,7 @@
       &:focus {
         background: color-mix(
           in srgb,
-          var(--gds-sys-color-blue) 50%,
```
splashdust, line 91: This one can be reverted

libs/core/src/components/segmented-control/segment/segment.trans.styles.scss
```diff
@@ -22,7 +22,7 @@
       appearance: none;
       background: transparent;
       border-width: 0;
-      color: var(--gds-sys-color-content-content);
```
splashdust, line 25: This one should be --text-primary-color for now, from Chlorophyll

splashdust (changes requested):

splashdust (approved):
👌

Issue #1806, opened 2024-11-18
Wizard: Icon has wrong alignment


akyataapp:
Wizard: Icon for in-page-wizards next button has wrong vertical alignment

### Bug already reported?

- [X] I confirm that I have checked if the bug already has been reported

### For which framework/library you are reporting the bug

React

### Component name

in-page-wizard

### Description

If you provide the optional icon to the next button, the alignment is wrong. The icon is showed a little bit to high. If you remove "height: 1.5rem;" at the line below, the problem is solved.
https://github.com/seb-oss/green/blob/d8c08002d1544923a63f5114fe411566891238b6/libs/chlorophyll/scss/components/wizard/in-page/_mixins.scss#L150

// Atay (S87875)

PR #1881, opened 2024-12-17
Wizard button icon alignment fix

EldRoos:
Also better storybook example text variations.
Now follows the design better on the icon center alignment.
Closes: #1806


EldRoos:
Got a report that this fix is not fixing the whole issue. Investigating!

PR #1882, opened 2024-12-17
Tooltip: Fix background color

maoq01:
Fix for styling for tooltip (based on tooltip styling from info-circle.component.scss).

Applied scss styling is declared in a separate `styles.ts` file.


vsjolander (approved):
LGTM

Issue #711, opened 2023-01-20
Labels have too much bottom margin on IBP


daneriseb:
This css leaks from IBP:

label {
    display: inline-block;
    margin-bottom: .5rem;
}

EldRoos:
Fixed in new core 2.

PR #1886, opened 2024-12-18
Reset label styling to fix leaking

EldRoos:
Closes: #711


splashdust (commented):
This change is potentially breaking, since it adds a element selector. Let's hold off on this one until the next major release, which should be fairly soon.

PR #1904, opened 2025-01-08
`gds-link` better `href` docs and make it optional

astrit:
…no empty string


astrit:
After further discussion and consideration, we decided not to allow the href attribute to be optional for the gds-link component. 

Omitting the href attribute can lead to significant accessibility issues, as it violates the Web Content Accessibility Guidelines (WCAG). Specifically, links without a valid href cannot be focused using keyboard navigation, which is essential for users who rely on assistive technologies.

Therefore, I will retract this PR.

PR #1905, opened 2025-01-08
Fix/component exports

maoq01:
## Fixes
- module exports
- toast module: removed `BrowserAnimationModule` (only allowed in AppRoot).
- input-mask fixes (to trigger new mask on dateLocale changes


libs/angular/src/v-angular/v-angular.module.ts
```diff
@@ -8,38 +8,54 @@ import { NgvButtonModule } from './button'
...
 @NgModule({
   declarations: [],
   imports: [CommonModule],
   exports: [
+    NggCharacterCountdownDirectiveModule,
+    NggCoreWrapperModule,
+    NggSlugPipeModule,
+    NggvDragDropModule,
+    NggvInputMaskModule,
     NgvBaseControlValueAccessorModule,
```
pontus-github, line 39: Could we add NgvAlertModule here as well?
```suggestion
    NgvAlertModule,
    NgvBaseControlValueAccessorModule,
```

PR #1927, opened 2025-01-20
refactor/update-react-segmented-control-props

terrance456:
- Update `segmented-control` component props according to latest `green/core` component


splashdust (approved):
@terrance456 Looks good, but I added the following:
* Keep `segMinWidth` but assign it to `min-width`, to avoid breaking existing API
* Deprecate the manual wrappers in favor of the generated ones. No need to keep having to update this manually when the underlying component changesm since the Lit wrapper already handles it automatically.

PR #1928, opened 2025-01-20
Typeahead: Added emptyOptionLabel Typeahead input

maoq01:
## Fix

- Added missing input for typeahead directive `emptyOptionLabel`.
- Update position for [hidden].
- Error converting null to object.


vsjolander (approved):
LGTM!

PR #1936, opened 2025-01-23
bugfix/safe-guard-tokenize-args

terrance456:
- Fixes error in react segmented control component

{50D1BCCD-04B2-42A7-AF3A-4E44765361F9}


splashdust (approved):
Looks good to me. Nice catch!

PR #1942, opened 2025-01-23
GdsText: Inherit from GdsContainer

splashdust:
Instead of redefining certain props on `GdsText`, I think it makes sense to inherit all the base props from `GdsContainer` instead. Although this adds a bit more overhead to `GdsText`, it is more performant than being forced to wrap it in a container.


vsjolander (approved):
I agree! Makes more sense!

PR #1944, opened 2025-01-23
Transform text tokens to `rem` in CSS variables

splashdust:
Other spacing tokens are still outputted as `px`


splashdust:
> Look good to me! Although we should definitely consider the spacing tokens to update them to rem as well. 

In general, paddings, margins, gaps, etc, should not scale with font size, so I think we'll need a more granular solution for spacing values. Perhaps some way of specifying token unit in style expression props.

PR #1946, opened 2025-01-27
bugfix/segmented-control-arrows-not-hidden

terrance456:
#### Bug: Arrows not hidden when Segment child changes

Initial segments having long label child

{93B35581-F0F0-4EEF-848E-DCBC9839A7ED}

After segment child changes arrow remains visible even though there is no segment available in the list.

{8F7CC17F-AAA2-476A-BDEA-291F6B643D6E}

### DIFF
--- a/libs/react/src/lib/segmented-control/segmented-control.tsx
+++ b/libs/react/src/lib/segmented-control/segmented-control.tsx
@@ -36,10 +36,10 @@ export interface SegmentProps {
   disabled?: boolean
   children?: ReactNode | ReactNode[]
   width?: string
-  'min-Width'?: string
-  'max-Width'?: string
+  minWidth?: string
+  maxWidth?: string
   /** @deprecated */
-  segMinWidth?: number
+  segMinWidth?: string
 }
 
 /** @deprecated - Use `GdsSegmentedControl` instead */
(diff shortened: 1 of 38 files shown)


libs/react/src/lib/segmented-control/segmented-control.tsx
```diff
@@ -36,10 +36,10 @@ export interface SegmentProps {
...
   children?: ReactNode | ReactNode[]
   width?: string
-  'min-Width'?: string
-  'max-Width'?: string
+  minWidth?: string
+  maxWidth?: string
   /** @deprecated */
-  segMinWidth?: number
+  segMinWidth?: string
 }
```
splashdust, line 43: I'm guessing this change is because you need to specify `px` for this to work now, but this is a breaking change to an already deprecated property. The point of deprecating it in the first place was to avoid breaking changes, so t would be better to just add `px` to it in the fallback on line 69.

The changes to minWidth/maxWidth are technically breaking too, but since they were just added and the component itself is deprecated, that change is unlikely to cause trouble, so I think we can keep that.

splashdust (changes requested):

libs/react/src/lib/segmented-control/segmented-control.tsx
terrance456: @splashdust fixed.

splashdust (approved):
Looks good! Thank you for this fix! 💚

PR #1955, opened 2025-01-30
Dropdown: Size attribute applied to gds-form-control-header

rostigcykel:
Added same logic that is already applied to gds-form-control-footer in Dropdown.

Also same that is found inside Datepicker component when applying size attribute to gds-form-control-header through a css class 

One discrepancy between datepicker and dropdown is that datepicker has size small/large and dropdown only small/medium but only small has specific css class inside gds-form-control-header so it seems to be OK.

Added tests for both gds-form-control-header and gds-form-control-footer cases

### DIFF
--- a/.changeset/flat-cooks-look.md
+++ b/.changeset/flat-cooks-look.md
@@ -1,0 +1,5 @@
+---
+'@sebgroup/green-core': patch
+---
+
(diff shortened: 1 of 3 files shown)


splashdust (approved):
Excellent, thank you for this fix! 💚

PR #1957, opened 2025-01-31
Feat/mabff 993 add grouping v angular table

sarjar:
Required functionality:
image


libs/angular/src/v-angular/table/table.component.scss
```diff
@@ -11,13 +11,14 @@
   --sg-table-sort-icon-color: var(--gds-ref-pallet-base600);
   --gds-ref-pallet-base500: hsl(0, 0%, 68%);
   --sg-table-sort-icon-color-active: #fff;
+  --sg-table-row-background-color: #f8f8f8;
```
splashdust, line 14: Here, I would recommend getting the value from an existing color variable instead:
```css
--sg-table-row-background-color: var(--sg-grey-100)
```
We will refactor the color variables setup in Chlorophyll soon:ish, but then it is good to be able to do a search and find all the places where they are used.

splashdust (changes requested):

PR #1966, opened 2025-02-05
Declarative Layout refine and refactor

splashdust:
## What is it
Since we've been trying out and using this system for a while now, some patterns and opportunities for improvement have started to emerge. This PR is a step towards finalizing the declarative layout system.

The focus of this PR is to streamline the available properties for building layouts using GDS tokens. That means that rarely used and non-layout related properties are being removed. Also ome support for custom values are being removed, except for sizing and positioning. Declarative Layout is not meant to replace CSS, just to reduce or eliminate the need for layout CSS.

#### Refactor
- Colors now only support tokens from Green, no custom values except `transparent` and `currentColor`. This simplifies implementation and makes color properties more consistent.
- Color properties now always use `tokenName/transparency`, where transparency is optional.
- Set `box-sizing` to `border-box` by default
- Set `border-color` to `primary` by default
- `border` property now works like css shorthand border property, ie. `size style color`
- `border-style` property added for completeness
    - We now have `border`, `border-width`, `border-color`, `border-style`, and  `border-radius`, all with the same parameter structure as css.
- Commonly used `@styleExpressionProperty` logic moved to reusable functions.
- `gds-grid`: `auto-columns` now support any CSS units. If no unit is specified, it still appends `px` for backwards compatibility.

#### Removed
These properties have been removed. Any edge-cases that need these will have to be written in regular CSS

- `pointer-events`
- `user-select`
- `cursor`
- `transform-style`
- Animation related properties. These we could eventually handle through a well thought out animation/transition system instead
    - transition
    - transition-behavior
    - animation

#### Other improvements
- Composable and reusable Style Expression Property groups through class mixins
  - This means that other components can now have groups of style expression properties without adding much to overall bundle size. Buttons, form components, filter chips, segmented control, etc, now support common properties such as flex/grid item props, margins and horizontal sizing.
- Put Declarative layout properties in their own section in the API tables. This makes it clear which properties are specific to the component, and which are just styling.
- Refactored default values from properties to CSS. Setting property defaults during initialization adds some scripting overhead. Moving these to CSS improves base performance slightly.

#### Other changes in this PR
- Add role attribute to `gds-divider`
- Fix focus outline in calendar (2023 style)

#### Deprecated
- `gds-container` is deprecated in favor of `gds-div`
- `gds-spacer` is deprecated in favor of `gds-div`


libs/core/src/components/card/card.ts
```diff
@@ -69,9 +71,8 @@ export class GdsCard extends GdsContainer {
     super()
 
     // Default values for cards
-    this.padding = 's; m{l}'
-    this['border-radius'] = 'xs; m{s}'
-    this['border-width'] = '4xs'
+    this.padding = 's;m{l}'
+    this['border-radius'] = 'xs;m{s}'
```
astrit, line 75: I would suggest here that we add as default display to `flex` reason for that is most times we are having to wrap again content to do that, most common is flex-direction column. 

```ts
this.display = 'flex'
this['flex-direction'] = 'column'
this['gap'] = 's;m{l}'
```

libs/core/src/components/card/card.ts
```diff
@@ -6,35 +6,38 @@ import {
...
   ]
 
   /**
-   * Controls the box-shadow property.
+   * Style Expression Property for the `box-shadow` property.
    *
-   * These are the available values you can use to define shadow size
+   * Accepts shadow tokens from the design system.
    *
    * `xs`, `s`, `m`, `l`, `xl`
```
astrit, line 42: The shadows seem to not be working!

astrit (changes requested):
Super good update! 
The renaming to `gds-div` makes so much more sense now. 

I would suggest some minor changes on the `gds-card` specifically. 
Adding flex as default with column direction to reduce number of flexes needed which I think is the most common use case. 

Also the shadow doesn't seem to work. 

And perhaps we should update exmaples across the storybook since some seem to be broken let me know if you need help with that! 

Other than that super great update!

libs/core/src/components/card/card.ts
splashdust: Good catch, I agree that flex makes good sense as a default for card!

libs/core/src/components/card/card.ts
splashdust: The shadow property is still only available on `gds-card`.

Originally we said that shadows should only ever be used on cards, but perhaps it would be best to move that one up to `gds-div` as well, for consistency. What do you think?

splashdust: Ok, here's what I did:
- Add `box-shadow` to `gds-div`, to keep in line with "same names as css"
- Keep `shadow` on `gds-card` for now, but tag as deprecated

astrit (approved):

PR #1984, opened 2025-02-19
New component: Radio Button

splashdust:


splashdust:
@astrit 

Superb work as always. The interaction experience and animations feels top notch! 👌 

Here's what I found that we need to fix before merge:

- [ ] When using VoiceOver, `ctrl-opt-space` should mark the radio as selected, but it's not working. Not sure exactly why this happens, but probably has something to do with events and where they originate. I think VoiceOver simulates a click-event, but it may be on a different DOM level than a normal click. Something like that.
- [ ] Groups are missing accessible labels:
   image
- [ ] Change events dispatched from `gds-radio` should probably not bubble, since that will cause events listeners on `gds-radio-group` to get called twice (once for the radio and once for the group)
- [ ] Disabled controls should not be `inert`, as this completely takes them out of the a11y tree. I noticed we have this issue on other controls as well, and it seem to stem from `<gds-field-base>`. If you prefer, we can leave it as is for this PR and open a new one to fix in all controls.

astrit (approved):

PR #1985, opened 2025-02-19
Added spinner component 

yuyo2:


libs/core/src/components/spinner/gds-spinner.ts
```diff
@@ -0,0 +1,173 @@
...
+   * @type {string|undefined}
+   */
+  @property({ type: String })
+  text?: string
+
+  /**
+   * Controls text visibility mode.
+   * When true, text is visible and announced to screen readers.
+   * When false, text is only announced to screen readers.
+   * @type {boolean}
```
splashdust, line 68: Type information is not needed in doc comments, since typescript can already infer this from the code.

libs/core/src/components/spinner/gds-spinner.ts
```diff
@@ -0,0 +1,173 @@
...
+    `,
+    styles,
+  ]
+
+  /**
+   * Loading text to be displayed. Can be shown visually or for screen readers only.
+   * @type {string|undefined}
+   */
+  @property({ type: String })
+  text?: string
```
splashdust, line 62: I think we should rename this to `label` for consistency with other components.

libs/core/src/components/spinner/gds-spinner.ts
```diff
@@ -0,0 +1,173 @@
...
+
+  /**
+   * Controls text visibility mode.
+   * When true, text is visible and announced to screen readers.
+   * When false, text is only announced to screen readers.
+   * @type {boolean}
+   * @default true
+   */
+  @property({ type: Boolean })
+  visualText: boolean = true
```
splashdust, line 72: Here, I would invert this boolean and call it `hideLabel` instead. Then it plays nicer with how boolean HTML attributes work, where true is implied when the attribute is present, and false otherwise:

```html
<gds-spinner label="Loading..." hide-label></gds-spinner>
```

libs/core/src/components/spinner/gds-spinner.ts
```diff
@@ -0,0 +1,173 @@
...
+    this.setAttribute('aria-busy', 'true')
+    TransitionalStyles.instance.apply(this, 'gds-spinner')
+  }
+
+  /**
+   * Generates CSS classes for the wrapper element based on component state.
+   * @private
+   * @return {string} Space-separated class names
+   */
+  private _getWrapperClasses(): string {
```
splashdust, line 115: Here, I would return a object that maps the booleans, which can be fed to Lit's `[classMap](https://lit.dev/docs/templates/directives/#classmap)` directive.

Also, by convention, we use native JS private field syntax were possible, ie `#field` rather than `private _field`. [Here's the rationale.](http://localhost:4400/?path=/docs/contributing-coding-guidelines--docs#-for-private-properties)

I.e, something like
```ts
#getWrapperClasses = () => ({
  'gds-spinner-fullscreen': this.fullscreen,
 ...etc
})
```

libs/core/src/components/spinner/gds-spinner.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    return classes.join(' ')
+  }
+
+  /**
+   * Generates CSS classes for the spinner element based on size.
+   * @private
+   * @return {string} Space-separated class names
+   */
+  private _getSpinnerClasses(): string {
```
splashdust, line 138: Same here

libs/core/src/components/spinner/gds-spinner.ts
```diff
@@ -0,0 +1,173 @@
...
+  size: 'default' | 'md' | 'lg' | 'sm' = 'default'
+
+  /**
+   * Lifecycle method called when component is added to DOM.
+   * Sets up accessibility attributes and applies transition styles.
+   * @return {void}
+   */
+  connectedCallback(): void {
+    super.connectedCallback()
+    this.setAttribute('role', 'status')
```
splashdust, line 105: In terms of semantics I think 'progressbar' is a better suited role. Then we can also extend it in the future to convey quantifiable progress. Also, `aria-busy` should be added to the region which will be updated when loading is done. To keep it simple, I think we'll need to leave that to the consumer for now.

Here is an example of indeterminate progress using the `<progress>` element:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress#describing_a_particular_region

Maybe down the road we can add a parent component, something like `<gds-progress>`, that can wrap a loading region to add a spinner or progress bar and automatically wire everything up with correct aria roles.

splashdust (changes requested):
Overall, super good work!

Did find some things to nit pick about.
Hojta on teams if you want to bolla :)

libs/core/src/components/spinner/gds-spinner.ts
yuyo2: Status vs. Progressbar for Spinners
Using role="status":
Perfectly valid for spinners that don't show measurable progress
Simpler implementation with fewer required attributes
Communicates to screen readers: "Here's a status update about loading"
Many design systems and accessibility experts recommend this for indeterminate spinners
Using role="progressbar" with undefined aria-valuenow:
More semantically tied to the concept of "progress in process"
Requires additional attributes (aria-valuemin, aria-valuemax)
Some screen readers may announce it differently (as a progress indicator)
Better suited if your spinner might eventually display determinate progress
The WAI-ARIA authoring practices actually support both approaches for indeterminate spinners.

splashdust: Allright, I buy this argument. Let's leave it!

patrikfogelstrand:
Can we assist in any way to get this merged and released? @splashdust

splashdust (approved):

PR #1986, opened 2025-02-20
Checkbox & Radio hover styles

vsjolander:
This PR alters the current 2016 checkbox & radio hover styles and adds our 2016 color tokens.

There's an issue on mobile where the hover styles make it look like multiple items are selected. 

We need to update Figma aswell.


splashdust (approved):
Looks good to me!

PR #1989, opened 2025-02-20
Add translations for more languages

splashdust:
Also in this PR: Add a language selector in Storybook to set component language.


vsjolander (approved):
Looks good!

Maybe we should involve @Josie-O in this aswell. I can image she would like to have a look at it!

astrit (approved):

PR #2005, opened 2025-02-28
fix: added actionEmitter and close on option click on foldout

JohanAtSEB:
**Angular**
Added `actionEmitter` to emit an empty value when the foldout closes. This is to make it easier to use the foldout if a component has `ChangeDetectionStrategy.OnPush`. Do a detectChanges() when the foldout emits.

Changed that the foldout closes when you click on an option. Previously, it only closes on space or enter.

### DIFF
--- a/libs/angular/src/v-angular/modal/fold-out/fold-out.component.ts
+++ b/libs/angular/src/v-angular/modal/fold-out/fold-out.component.ts
@@ -12,14 +12,8 @@ import {
   ViewChild,
 } from '@angular/core'
 
-import {
-  fromEvent,
-  Subscription,
-} from 'rxjs'
-import {
-  filter,
-  takeWhile,
-} from 'rxjs/operators'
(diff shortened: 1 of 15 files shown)


libs/angular/src/v-angular/modal/fold-out/fold-out.component.ts
```diff
@@ -1,18 +1,19 @@
...
+  EventEmitter,
   HostBinding,
-  HostListener,
   Input,
   OnDestroy,
+  Output,
   ViewChild,
 } from '@angular/core'
 
-import '@sebgroup/green-core/components/icon/icons/dot-grid-one-horizontal.js'
```
vsjolander, line 12: This should not be removed. It's the web component for the icon.

vsjolander (changes requested):

vsjolander (approved):
LGTM

PR #2008, opened 2025-03-03
Input: Adding validation icon

vsjolander:


splashdust (changes requested):
The icon should have the same color as the error text, but currently it gets the default text color:
image

splashdust (approved):

Issue #1978, opened 2025-02-17
Form components: Do not look correct when invalid.


VaidasMasys:
### Bug already reported?

- [x] I confirm that I have checked if the bug already has been reported

### For which framework/library you are reporting the bug

Angular

### Component name

dropdown, input, textarea, datepicker

### Description

Once field becomes invalid look of invalid form field does not meet criteria. Bottom borders for an invalid field should become red

### Steps To Reproduce

1. Make every type of form component (inputs, datepickers ,dropdowns and so on) invalid.

### Current Behaviour

![Image](https://github.com/user-attachments/assets/8c5564c0-4664-46e7-b202-316775f54b33)

### Expected Behaviour

![Image](https://github.com/user-attachments/assets/454aafa1-7c23-420d-ba2e-01639f83838d)

PR #2019, opened 2025-03-05
fix: added red line to input fields when there's an error. made slide-out scrollable. exposed interfaces

JohanAtSEB:
Added `.nggv-field--error` class in `input.component.scss` and `textarea.component.scss` to be able to showcase when an error has occurred. Also to match other components, such as dropdown. Fix for https://github.com/seb-oss/green/issues/1978

Made `slide-out` scrollable when needed

Added map `core` to a `index.ts` file to expose some interfaces, `Option`, `OptionGroup`, `OptionBase` etc.


libs/angular/src/v-angular/input/input.component.scss
```diff
@@ -45,6 +45,12 @@
       @include validation.add-feedback(tokens.$intent-danger-background);
     }
 
+    &.nggv-field--error {
+      $_border-bottom: 0.125rem solid #9f000a !important;
+      border-bottom: $_border-bottom;
+      height: 100%;
```
JohanAtSEB, line 51: `height: 100%` needed, otherwise the border would default back to it's original size (half of this).

vsjolander (approved):

PR #2056, opened 2025-03-17
New component: Checkbox

astrit:
This PR adds the new Checkbox component, and adresses some other adjacent concerns that came up along the way:

- Adopt a new pattern for sharing internal styles and DOM structures.
- Refactor Radio buttons to adopt this pattern and share styles with Checkbox.
- Add `NggCoreCheckbox` directive for Angular for easy integration with Reactive Forms.
- Fix various validation and state related issues across the board, which came up during testing.

---

# **Technical Requirement Specification - Checkbox**

### **1.1 Component Name**

Name: Checkbox

Elements: `<gds-checkbox-group>` and `<gds-checkbox>`

### **1.2 Description and Purpose**

A checkbox is a boolean control used to toggle a single option on or off. Unlike radio buttons, multiple checkboxes can be selected independently. Checkboxes are commonly used in forms, settings, and filters.

A **checkbox group** is a form control that manages multiple checkboxes as a set. Its value is an **array of selected values**. This component improves accessibility and form-handling by simplifying value management.

### **1.3 Scope**

The checkbox and checkbox group shall be developed and tested as form controls. Other potential use cases are outside the scope.
 
### **1.5 References**

- https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role
- https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/
- https://github.com/shoelace-style/shoelace/tree/next/src/components/checkbox

### **1.6 Variants**

- **Default**  
  - The standard checkbox with a label and optional supporting text.
- **Contained**  
  - A checkbox inside a rounded primary gray background. The contained version will not display label support text.

### **1.7 States**

- **Checked**: The checkbox is selected.
- **Unchecked**: The checkbox is not selected.
- **Indeterminate**: The checkbox is in a mixed state (e.g., when only some sub-options are selected). This is a visual state only and does not affect the actual value.

### **1.8 Dependencies**

- The checkbox component shall extend the `GdsFormControlElement` base class.
- The checkbox group component shall make use of `<gds-form-control-header>` and `<gds-form-control-footer>`.

---

### **Anatomy**

A **checkbox group** shall have an anatomy roughly as follows:

```
- flex column
    - form control header
    - main slot for checkboxes
    - form control footer
```

A **checkbox** shall have an anatomy roughly as follows:

```
- flex row
    - square checkbox element
    - flex column
        - label
        - label supporting text
```

### **2.3 Accessibility Requirements**

- **Checkbox Component**  
  - The checkbox shall mimic `<input type="checkbox">` behavior and implement the `checkbox` role.  
    - [https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox)  
    - [https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/checkbox_role)  
  - The checkbox shall support the **indeterminate** state when applicable.
  - Keyboard navigation:  
    - **Spacebar** toggles the checkbox.

- **Checkbox Group Component**  
  - The checkbox group shall implement the `group` ARIA role.  
    - [https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/](https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/)  
  - The checkbox group shall maintain an **array of selected values**.
  - Keyboard navigation:  
    - **Tab** moves focus between checkboxes.
    - **Spacebar** toggles the focused checkbox.

---

### **3.1 Props/Attributes**

Both **checkbox** and **checkbox group** shall inherit all relevant properties from `GdsFormControlElement`.  

#### **Additional Checkbox-Specific Properties**
| Name          | Type      | Default  | Description |
|--------------|----------|----------|-------------|
| `checked`     | `boolean` | `false`  | Whether the checkbox is checked. |
| `indeterminate` | `boolean` | `false` | Whether the checkbox is in an indeterminate state. |

### **3.2 Events & Emissions**
| Event Name | Trigger         | Payload       | Description        |
|------------|---------------|-------------|--------------------|
| `change` | When a checkbox is toggled | `{ checked: boolean, value: string }` | Emitted when a checkbox changes state. |
| `input` | When a checkbox value is updated | `{ value: Array<string> }` | Emitted when the checkbox group updates its selected values. |

---

## **4. Implementation Details**

Since checkboxes are individual elements that does not need to be associated by name like radio buttons, using `<input type="checkbox">` internally may provide the most accessible solution.

For inspiration, take a look at Shoelace's `checkbox` implementation:  
[https://github.com/shoelace-style/shoelace/tree/next/src/components/checkbox](https://github.com/shoelace-style/shoelace/tree/next/src/components/checkbox)

As mentioned, this component shall inherit from `GdsFormControlElement` and follow the same conventions for form control validation and behavior as the other form controls in Green.

The primary design goal for form components in Green is to mimic their native counterparts as far as practically possible and adhere to standard ARIA roles and guidelines.

The element names shall also be added to the `NggFormControlsModule` in `@sebgroup/green-angular` to ensure compatibility with Angular Reactive Forms.

Internal layout structure shall not use Green declarative layout. For performance reasons, it is better to use regular CSS internally in components.


splashdust:
@astrit Can't select you for review, since you opened the PR, but please have a look :)

vsjolander (approved):
Wooopwoop 🥳🎉 this is amazing work. Many people will be pleased with the addition of this component. Well done @splashdust and @astrit 💎

astrit:
Yayyy, super great work @splashdust  finally we have a checkbox 🦾

Super nice that you separated the primitive and that R to the B to the C to the B style template 😂

I have some comments: 
I added a new inderterminate state not completely sure if it should be handled this way and what scenarios should be used since I guess when "Select All" is one of the options. 

Please review this commit [821bf79](https://github.com/seb-oss/green/pull/2056/commits/821bf79a49e28d8e8c7c9f07dc2a1277b739edb8)

Also on design the check disabled style should probably be removed. 

Dropdown: 
Should we change now and deploy at the same time to use the checkbox there or after this merge ?

PR #2085, opened 2025-03-25
SlideOut: Focus on the firstFocusable element

HappyCtrl:
## Fix
Set focus on the `firstFocusable` element instead of the `lastFocusable` element.
### Previous behaviour
If the slideout contains content that causes a scroll, the slideout would be opened and scrolled to the bottom, so the `lastFocusable` element became visible in the viewport.
### Proposed new behaviour
The `firstFocusable` element should be visible upon opening the slideout.


HappyCtrl:
Would it be possible to get a review on this PR? 😊
Ping @vsjolander @splashdust

vsjolander (approved):

PR #2086, opened 2025-03-26
Docs: Changed default port for chlorophyll & charts storybook

EldRoos:
Chlorophyll & charts gets port 4403, 4404 & 4405.
Port 4400, 4401 & 4402 is taken by core, react and angular.


splashdust (approved):
👍

Issue #1990, opened 2025-02-21
Dropdown: Screen-reader reads chevron-down


DovydasMar:
### Bug already reported?

- [x] I confirm that I have checked if the bug already has been reported

### For which framework/library you are reporting the bug

Angular

### Component name

dropdown

### Description

when you open screen-reader and hover over dropdown it says "chevron down", I spoke with CX person, he said that it's not supposed to say that

![Image](https://github.com/user-attachments/assets/bcee111e-8eef-4247-88b7-5007c49f4c34)

### Steps To Reproduce

1. using NVDA screen reader.
2. focus on dropdown.
3. it says everything you need it to say and on top of that it says "chevron-down" 

### Current Behaviour

it says "chevron-down"

### Expected Behaviour

it should say something like "click to expand" (discuss with your design team)

PR #2089, opened 2025-03-27
Dropdown custom expand icon label

EldRoos:
I defined the default label to **Expand**, but maybe we should choose something like **Open** or something else?

Closes: #1990


libs/core/src/components/dropdown/dropdown.ts
```diff
@@ -293,7 +300,10 @@ export class GdsDropdown<ValueT = any>
           ${this.combobox && !this.multiple
             ? this.#renderCombobox()
             : this.#renderTriggerButton()}
-          <gds-icon-chevron-bottom slot="trail"></gds-icon-chevron-bottom>
+          <gds-icon-chevron-bottom
+            slot="trail"
+            label=${this.expandLabel}
```
splashdust, line 305: We can hard-code the icon label here instead, but make sure to make it localizable with `msg`:
```
label=${msg('Expand')}
```

Then you need to add the translations:
1. Enter the core directory: `$ cd libs/core`
2. Extract translation strings: `$ npx lit-localize extract`
3. Edit the language files in `xliff/*.xlf` and update the added entry for "Expand"
4. Generate updated typescript locale files: `$ npx lit-localize build`

Commit all the changes, including the generated files.

Read more here: https://lit.dev/docs/localization/overview/#extracting-messages

splashdust (changes requested):

splashdust (approved):
👍

PR #2095, opened 2025-03-31
Update input.styles.ts

vsjolander:


astrit (approved):
Fantastic job as always, tried it out pretty compact 👌

PR #2098, opened 2025-03-31
Improvements: Link, Text and Div

astrit:


libs/core/src/components/text/text.style.ts
```diff
@@ -2,7 +2,7 @@ import { css } from 'lit'
 
 const style = css`
   :host {
-    display: block;
+    display: inline-block;
```
splashdust, line 5: I think we should leave this as is for now, since it would be a layout-breaking change for current consumers, and it can still be controlled using the display property if needed.

Down the line it could make sense to have different defaults depending on tag name, ie headings and paragraphs as `block`, span as `inline`, etc.

splashdust (changes requested):

splashdust (approved):
💯

Issue #1221, opened 2024-02-19
Datepicker: Allow to hide clear/today buttons


nerjok:
### Description

As per team product owner, designer, we express a wish to have ability not to show clear button to simplify calendar usage.

### Who will benefit from this feature?

user is not allowed to delete date and would be forced to select/change  without showing additional validations

EldRoos:
Fixed in https://github.com/seb-oss/green/releases/tag/%40sebgroup%2Fgreen-core%401.78.0

Issue #2105, opened 2025-04-01
Datepicker: Optional setting for the Clear button


DagFred:
### Description

It should be optional to be able to choose whether you want a "Clear" button or not.

### Who will benefit from this feature?

If the datepicker is a required field, then there shouldn't be a "Clear" button. The only purpose of the "Clear" button would be to create an error state, since the datepicker shouldn't be left empty in this case. Therefore it is very misleading to show the "Clear" button to the user and it will create a poor user experience.

EldRoos:
Duplicate of #1221.
Will try to fix this soon. Already made the changes a while ago, just missing the unit tests.

splashdust:
Fixed in https://github.com/seb-oss/green/releases/tag/%40sebgroup%2Fgreen-core%401.78.0

PR #2108, opened 2025-04-02
Datepicker: Option to hide clear and today

EldRoos:
Need to add tests.

Closes: #1221 #2105


libs/core/src/components/datepicker/datepicker.component.ts
```diff
@@ -103,6 +103,18 @@ class Datepicker extends GdsFormControlElement<Date> {
   @property({ type: Boolean, attribute: 'hide-label' })
   hideLabel = false
 
+  /**
+   * Whether to hide the Clear button under the calendar.
+   */
+  @property({ type: Boolean, attribute: 'hide-clear' })
+  hideClear = false
```
splashdust, line 110: @EldRoos I know you're still in draft-mode, but front-running this anyway :)

I think this one should be called `clearable` to align with other components, and thereby also default to `false`. In the future, maybe this should also be moved to a clear-button in the field like in other components as well, but for now, let's align the API-naming at least.

splashdust (approved):
Sorry for the delayed review on this one.
Looks good to me now. Great work! 💚

PR #2112, opened 2025-04-03
Green senior

vsjolander:


vsjolander:
Doing this in https://github.com/seb-oss/green/pull/2151

PR #2114, opened 2025-04-04
Formatting and sensitive data

Wikenbrant:


astrit:
Super good contribution so far, I like it 👌

Since is still a draft and if you think is a good idea perhaps extending the `gds-text` instead it could be more efficient ? 

I can imagine this to be used with different sizes and styles, for example: 

```html 
<gds-text format="number" font-size="detail-s">1420</gds-text>
<gds-text format="date" font-size="detail-l">...</gds-text>
```

Wikenbrant:
> Super good contribution so far, I like it 👌
> 
> Since is still a draft and if you think is a good idea perhaps extending the `gds-text` instead it could be more efficient ?
> 
> I can imagine this to be used with different sizes and styles, for example:
> 
> ```
> <gds-text format="number" font-size="detail-s">1420</gds-text>
> <gds-text format="date" font-size="detail-l">...</gds-text>
> ```

they all extend gds-text 

```
abstract class GdsFormattedText extends GdsText {
  abstract get formattedValue(): string

  @query('[tag]')
  private element?: Element

  protected updated(changedProperties: PropertyValues): void {
    super.updated(changedProperties)

    if (this.element) {
      this.element.innerHTML = this.formattedValue
    }
  }
}
```

libs/core/src/components/formatted-text/account/formatted-account.component.ts
```diff
@@ -0,0 +1,38 @@
...
+ * `gds-formatted-account` extends `gds-text` and formats the account to the desired format.
+ *
+ * @example
+ * <gds-formatted-account
+ *   account="54400023423">
+ * </gds-formatted-account>
+ */
+@gdsCustomElement('gds-formatted-account')
+export class GdsFormattedAccount extends GdsFormattedText {
+  @property()
```
splashdust, line 21: Here it would be good with a doc comment explaining the property, perhaps also adding that 11 digits are expected.

splashdust (changes requested):
This looks like a great way to structure a composable collection of utility components for formatting. Super excellent stuff! 

Can't really think of anything I would want to change here, other than you can remove the todo comments about logging/throwing (see my inline comment), and also it would be good to add DocComments to all the properties. Simple one-line explanation of what the prop does is usually enough. This will be displayed in the API-tables in Storybook.

libs/core/src/components/formatted-text/account/account-formatter.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export const accountsFormats = {
+  'seb-account': (value) => {
+    const account =
+      typeof value === 'string'
+        ? value.replace(' ', '')
+        : (value?.toString() ?? '')
+
+    if (account.length !== 11) {
+      // TODO: Should this throw an error?
```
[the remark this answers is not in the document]
Wikenbrant: fixed

libs/core/src/components/formatted-text/account/formatted-account.component.ts
Wikenbrant: fixed

Wikenbrant:
> This looks like a great way to structure a composable collection of utility components for formatting. Super excellent stuff!
> 
> Can't really think of anything I would want to change here, other than you can remove the todo comments about logging/throwing (see my inline comment), and also it would be good to add DocComments to all the properties. Simple one-line explanation of what the prop does is usually enough. This will be displayed in the API-tables in Storybook.
> 
> Oh, and also add a [changeset](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md)!
> 
> ```shell
> $ yarn changeset
> ```
> 
> Select patch level

added change set

splashdust (approved):
Everything looks good to me now. Fantastic work, and thank you for contributing to Green! 💚

PR #2116, opened 2025-04-07
Chlorophyll 4

vsjolander:


Maximilian-R (commented):
https://github.com/seb-oss/green/blob/65f9296e130f9b2ae7f5e186e1749467c2f361de/libs/chlorophyll/src/scss/components/skeleton-loader/_mixins.scss#L4

.skeleton-loader -> .gds-skeleton-loader ?

PR #2121, opened 2025-04-08
Fixing the test build paths for windows

EldRoos:
Need to test on mac/wsl.


splashdust:
Works on Mac and CI, let's keep it!

PR #2144, opened 2025-04-16
Doc: Updating a11y links

EldRoos:
Links like this will be on the general a11y page later on. No need to be repeated on each component.


vsjolander (approved):

astrit:
This PR depends on: #2156

PR #2145, opened 2025-04-16
Link: Add `aria-label`

astrit:


libs/core/src/components/link/link.component.ts
```diff
@@ -55,6 +54,12 @@ export class GdsLink extends withMarginProps(
   @property()
   rel?: string
 
+  /**
+   * The label of the link.
```
splashdust, line 58: Add some more info here:

Use this to add an accessible label, for example when only an icon is used in the link, or if different copy is needed for screen reader users.

splashdust (changes requested):

splashdust (approved):

PR #2151, opened 2025-04-16
Migrating to style-dictionary 4

vsjolander:
This pull request introduces a migration to Style Dictionary v4, along with a comprehensive rewrite of the token structure and output for better scalability. It also updates the design token references across multiple components to use the `--gds-sys-*` namespace, aligning the token structure.

The way tokens are built is now based on a themes and color schemes making easy to more of them. This also makes it possible to add density modes easily. More info can be found [here](https://github.com/amzn/style-dictionary/issues/1214#issuecomment-2128093553) from the maintainer of Style Dictionary.

Also did som minor fixes to Storybook.


astrit:
Super good job, can't wait to have this also out the door, super nice the density mode addition also 👌

Although I had a suggestion to remove the `--gds-sys-` prefix from variables for several reasons. 
I think we had this discussion before but could not find any documented decision. 

In my opinion the `-sys-` segment doesn't add functional value since we are using all tokens as system-level tokens meaning not really tailored to any component so practically all of them are system tokens. 

I looked into some other libs (including StyleDictionary it self) seems like most are not using the `-sys-` prefix. 
On top of that perhaps we could reduce code verbosity and smaller bundle size (though marginal 😄)

vsjolander:
It's up for debate but my take on it is:

I think we should have the token "[class](https://m3.material.io/foundations/design-tokens/overview)" in the name to know which token is being used. I think they add functional value. in the case if we need to trouble shoot later down the road it will be nice to have these to know what token you are looking at.

It also adds functionality when selecting tokens and visualizing how the system works.

Also if we start moving into Components tokens which we probably need to do to support the fine grained mods to themes, depending on naming convension names might clash. Having a separation there will simplify things in the long run.

The overhead it adds is not that much compared to the flexibility it gives.

@splashdust what's your take?

splashdust:
Explicit "token class" naming convention enables greater flexibility if/when we add other token classes in the future. On the other hand, it make all names longer, and they are already painfully long in some cases.

One alternative could be to say that `sys` is implied if no token class is specified (which I guess is the current state before this PR). This can create ambiguity though, and might lead to difficulties automating things and doing code analysis, etc.

In terms of overhead, it does add a bit. I counted the occurrences, and arrived at about 3.1kb extra just for the extra `-sys`. This is before gzip though, and since this is a repeating pattern most of it goes away, so it'll be very negligible in practice.

CSS variables can add quite a bit of bloat overall though, so in the long run, we may want to explore ways of minifying this at build time. Especially if we start to introduce component specific tokens.

splashdust (approved):
Looking good! I love how much cleaner the style dictionary setup feels now. Superb work! 💚

PR #2162, opened 2025-04-29
InputFields: Custom Locked Template

maoq01:
## Feature

Allow a custom template to be used for when the input (nggv-input, nggv-dropdown, nggv-textarea, nggv-date-input) is in a locked state.

If not sent in, `#defaultLockedTpl` will be used.

This in order to support below use-case (show a text together with the value of the input):
image

### How to use

Selected state is forwarded via `$implicit` context.

```html
<nggv-dateinput>
  <ng-template #lockedTpl let-selectedDate>Today ({{ selectedDate | date: 'shortDate' }})</ng-template>
</nggv-dateinput>

```

## Fix

Added toggle input for `small` in nggv-button story.


splashdust (approved):
Looks good to me!

PR #2173, opened 2025-05-05
Adding sizing tokens for transitional styles

vsjolander:
Issue can be seen here: https://storybook.seb.io/latest/angular/?path=/story/components-dropdown--select

Screenshot 2025-05-05 at 15 54 24


splashdust:
@vsjolander This shouldn't be needed, as the spacing tokens for 2023 should be present even when transitional styles is used. I think I found the underlying issue that caused this in #2174

PR #2178, opened 2025-05-07
Icon: Add color SEP

astrit:
# Add Direct Color Token Support to GDS Icons

## Overview
Adding native color style expression property to `gds-icon-*` components to enable direct color token application, eliminating the need for wrapper elements and improving developer experience.

## Problem
Currently, applying colors to icons requires either:
- Wrapping icons in `gds-div` or similar components
- Using inline style
- Targeting internal SVG parts with `::part(icon)` selectors

These approaches create unnecessary complexity and can lead to inconsistent implementations.

## Solution
Implement a `color` style expression property directly on `gds-icon-*` components that:
- Supports design system color tokens
- Handles transparency values

### Example Usage
```html
<!-- Before -->
<gds-div color="secondary">
  <gds-icon-bank></gds-icon-bank>
</gds-div>

<!-- After -->
<gds-icon-bank color="secondary"></gds-icon-bank>
```

If `color` proprty is not set it still defaults to `currentColor` 

CleanShot 2025-05-07 at 11 00 20@2x


splashdust (approved):
👌

PR #2181, opened 2025-05-08
New component Breadcrumb

astrit:
# Breadcrumbs Component

Adds a new breadcrumbs navigation component with support for custom separators, interactive elements, and proper accessibility.

## Key Features
- Supports both text-only and icon+text navigation links
- Responsive design with mobile optimization showing the previous page
- Configurable sizes (large/small)
- Each breadcrumb item is a slot that preferably uses `gds-link` component.

## Usage
```html
  <gds-breadcrumb>
    <gds-link href="/">
      <gds-icon-home-open slot="lead"></gds-icon-home-open>
      Home
    </gds-link>
  
  <gds-link href="/folder">
    <gds-icon-folder slot="lead"></gds-icon-folder>
    Documents
  </gds-link>
  
  <gds-link href="/folder/settings">
    <gds-icon-settings-gear slot="lead"></gds-icon-settings-gear>
    Settings
  </gds-link>
  
  <gds-text> Current page </gds-text>
  </gds-breadcrumb>
```

CleanShot 2025-05-15 at 13 39 07@2x


libs/core/src/components/breadcrumb/breadcrumb.component.ts
```diff
@@ -0,0 +1,59 @@
...
+  size: 'large' | 'small' = 'large'
+
+  render() {
+    const elements = Array.from(this.children)
+    const secondToLastIndex = elements.length - 2
+
+    return html`
+      <nav
+        role="navigation"
+        aria-label="Breadcrumb"
```
splashdust, line 34: I think it would be good to be able to control the label here.

A default value of "Breadcrumbs" does make sense, but it needs to be localized then. Check here for instructions: https://lit.dev/docs/localization/overview/#extracting-messages

libs/core/src/components/breadcrumb/breadcrumb.component.ts
```diff
@@ -0,0 +1,59 @@
...
+
+@gdsCustomElement('gds-breadcrumb', {
+  dependsOn: [IconChevronLeft],
+})
+export class GdsBreadcrumb extends withLayoutChildProps(
+  withSizeXProps(withMarginProps(GdsElement)),
+) {
+  static styles = [tokens, BreadcrumbCSS]
+
+  @property({ type: String })
```
splashdust, line 24: Missing doc comment on this property

libs/core/src/components/breadcrumb/breadcrumb.component.ts
```diff
@@ -0,0 +1,59 @@
...
+import { gdsCustomElement } from '../../utils/helpers/custom-element-scoping'
+import {
+  withLayoutChildProps,
+  withMarginProps,
+  withSizeXProps,
+} from '../../utils/mixins/declarative-layout-mixins'
+import { IconChevronLeft } from '../icon/icons/chevron-left.component'
+import BreadcrumbCSS from './breadcrumb.styles'
+
+@gdsCustomElement('gds-breadcrumb', {
```
splashdust, line 16: Missing doc comment for the class. I think this is also why no API-table is rendered in Storybook.

tsconfig.base.json
splashdust, line 1: Did you change anything in this file, or is it just formatting? If so, I think it would be good to revert, because otherwise prettier will change it back later and we'll have a new formatting diff.

splashdust (changes requested):
Found a few minor things, but overall super good work! 💯

tsconfig.base.json
astrit: Yes actually it was failing due to the order of packages that is why I changed it but for some reason the whole file got formatted to this style, will revert it!

libs/core/src/components/breadcrumb/breadcrumb.component.ts
astrit: Fixed it somehow i had removed the comment! 

Table also shows now 👌
CleanShot 2025-05-16 at 14 10 54@2x

libs/core/src/components/breadcrumb/breadcrumb.component.ts
astrit: Comment added in here ✌️

libs/core/src/components/breadcrumb/breadcrumb.component.ts
astrit: I added a new label prop so it's up to users to change that, default will be 'breadcrumb'

splashdust (approved):

PR #2201, opened 2025-05-27
Repo Tools New Component Small fix

EldRoos:
Wrong imports. Better doc.


libs/repo-tools/src/generators/core-component/files/__fileName__.stories.ts.template
```diff
@@ -25,7 +23,10 @@ const DefaultParams: Story = {
 }
 
 /**
- * TODO: Add documentation
+ * TODO: Add documentation here. Add the export in
+ * green/apps/docs/design/example/index.tsx and the beta
```
splashdust, line 27: This part is not related to storybook, so I don't think this is the right place to document this. The reminder about beta flag is good though!

splashdust (changes requested):

splashdust (approved):

PR #2226, opened 2025-06-12
 Alert component

yuyo2:


libs/core/src/components/alert/alert.stories.ts
```diff
@@ -0,0 +1,210 @@
...
+    buttonText: { control: 'text', name: 'Button Text (CTA)' },
+  },
+}
+
+export default meta
+
+type Story = StoryObj<GdsAlert>
+
+const Template = (args: any, { parameters }: any) => html`
+  <gds-div padding="xs; l {m}" margin="m">
```
patrikfogelstrand, line 31: I think the story should show exactly how the component renders, so remove this div to show the "reality"

splashdust (changes requested):
Great work so far! I highlighted a few thing that I think we should address before merging, but overall really good stuff!

Let me (or Ville) know if you need assistance with any of this, or don't have time or anything like that. I know getting a bunch of change requests can feel like a never ending story sometimes :)

libs/core/src/components/alert/alert.component.ts
```diff
@@ -0,0 +1,273 @@
...
+import '../icon/icons/triangle-exclamation.js'
+import '../icon/icons/circle-info.js'
+import '../icon/icons/cross-small.js'
+import '../button/index.js'
+
+import { tokens } from '../../tokens.style'
+import { alertStyles } from './alert.style'
+
+// Type definitions
+type AlertVariant =
```
patrikfogelstrand, line 22: @splashdust since these types are used a little bit everywhere throughout the components, are they typed somewhere else and can be inherited?

If not, I think these should be exported also @yuyo2

libs/core/src/components/alert/alert.component.ts
```diff
@@ -0,0 +1,273 @@
...
+
+  private _renderCloseButton() {
+    return this.dismissible
+      ? html`
+          <gds-button
+            class="close-btn"
+            variant="neutral"
+            rank="tertiary"
+            size="small"
+            aria-label="Dismiss alert"
```
patrikfogelstrand, line 217: Hard coded aria label won't work with translations, right?

libs/core/src/components/alert/alert.component.ts
```diff
@@ -0,0 +1,273 @@
...
+      : nothing
+  }
+
+  private _renderTimerBar() {
+    return this.timeOut > 0
+      ? html`
+          <div
+            class="timer-bar"
+            role="timer"
+            aria-label="Auto-dismiss timer"
```
patrikfogelstrand, line 232: Should this be hard coded?

libs/core/src/components/alert/alert.component.ts
```diff
@@ -0,0 +1,273 @@
...
+ * @fires close - Fired when alert is dismissed
+ * @fires action - Fired when action button is clicked
+ */
+@gdsCustomElement('gds-alert', {
+  dependsOn: [GdsButton, GdsCard],
+})
+export class GdsAlert extends GdsElement {
+  static styles = [tokens, alertStyles]
+
+  @property({ type: String, reflect: true })
```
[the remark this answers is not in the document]
yuyo2: For Lit properties, the String in the @property decorator (e.g., { type: String }) refers to the runtime type used for attribute conversion (HTML attribute to JS property). The TypeScript type (e.g., [variant: AlertVariant = 'information is for static type checking.

vsjolander:
@splashdust I've tweaked some stuff and I think this is good to be merged. Could you just scan one last time? <3

splashdust:
Tweaked a few more things:
- Add doc comments, for intellisense and API-tabel
- Remove additional hard-coded copy
- Removed the type exports. I prefer to avoid stuff like that if possible, since it adds unnecessary API surface, and requires naming things. If you need to get the type of variant, you can just use typeof instead. Also moved some of the type definitions inline, since they get spelled out in the API table and in intellisense hints that way.
- Change to javascript private field syntax where possible.
- I also removed dismissLabel and timeoutLabel, since I think it's enough to have built-in copy for these.
- Add beta tag

splashdust:
Now we just need to figure out what's up with the scss loader that prevents tests from running. I don't quite understand why it has happened in this PR, since it's seemingly unrelated, but needs to be solved before merge.

PR #2240, opened 2025-06-18
Alert: css correction

yuyo2:


astrit:
> A few thoughts:
> 
> * Setting max-width to `100vw` could be a bit unexpected, since it would prevent the alert from becoming wider than the viewport, even if its parent is. This can lead to some unexpected results when using zoom and similar situations.
> * Auto margin is also unexpected as a default. In fact, I think we should remove margin altogether here, and leave surrounding space up to the consumer. In practice, this could be gap, or padding from a parent, where this built-in margin would just get in the way.
> 
> Since these properties are on `:host`, a consumer can easily set it themselves if they need, so we can be conservative with the defaults here.
> 
> In conclusion, better defaults would be:
> 
> * Set max-width to 100% (and remove `width`, since with of 100% is already implied by display block)
> * Remove margin
> 
> ```css
> :host {
>     box-sizing: border-box;
>     min-width: fit-content;
>     max-width: 100%;
>     display: block;
>     container-type: inline-size;
> }
> ```

+1 to what Jocke said here. 

Few thoughts:
1. Let's consider updating this component soon to let users handle sizing instead
2. Maybe extend `Card` directly to simplify.
3. Could switch from grid to flex to conditionally show/hide the action area instead of leaving empty space (see screenshot).
4. Icons aren't rendering in React for some reason.
5. Focus outline is showing on both the button and the component.

CleanShot 2025-06-18 at 17 26 25@2x

CleanShot 2025-06-18 at 17 31 02@2x

CleanShot 2025-06-18 at 17 33 22@2x

splashdust (approved):

PR #2244, opened 2025-06-24
Green Core 2.0

vsjolander:
## Todo
- Write senior green card guideline.
- We need to handle global styles registry for tokens on root elements, should this be scoped aswell? We are using it for scroll lock on modals. (Updated in #2321)
- Add filter-chip removable variant
- - **Icon component**: will depricate few icons that should be removed with this release and also the sizing width and height property to be removed and communicate the change. 

## BREAKING CHANGES / Migration guide
- Removed disabled property on Badge in favor for making it a variant.
  - Migration step: use `<gds-badge variant="disabled">` instead of `<gds-badge disabled>`
- Badge: There is no positive variant of the notification property of the badge
  - Migration step: No action necessary, but variant will be ignored when the notification property is set to true.
- `font-size` changed to `font`, and font tokens updated.
  - Migration step: Change all occurences of the `font-size` style expression property and replace with font. Add font-weight to token name. Example: `<gds-text font-size="body-m">` --> `<gds-text font="body-regular-m">` 
- Colors; overhaul of color tokens
  - Migration step: Many of the color tokens used for `color`, `background`, `border-color`, and similar properties that accepts color tokens, have changed names. Refer to the color documentation page to find the updated names.
- Removed levels from Divider since it's no longer relevant due to border lot being a part of the levels structure.
  - Migration step: If you have been using the levels prop to control the color you should now insted use the color prop to set the possible colors on the divider. See color documentation to see possible border colors.
- Button variants `brand`, `warning` and `notice` is now added. Default button variant is now `neutral` instead of `brand` that was before.
   - Migration step: Add `variant="brand"` to buttons that should have the senior green colour. 
- Removed `shadow` prop from `gds-card`
   - Migration step: Use `box-shadow` property instead of `shadow`. 
- Removed `gds-container`
   - Migration step: Use `gds-div` instead. 
- Removed `gds-spacer`
   - Migration step: Use `gds-div` instead, and set height using the `height` property.
- Change default design-version of `gds-theme`
   - Migration step: To keep 2016 style, set the design-version attribute to "2016"
- `gds-theme` now sets the `color` CSS property by default. If you were relying on inheriting `color` from higher in the document, you need to override this now.
- Icon: Removed deprecated `width` and `height` properties.
   - Migration step: Use the `size` property instead. 

### Quick color-token migration cheat-sheet for the most commonly used colors

**Background**
- primary -> neutral-01
- secondary -> neutral-02
- tertiary -> neutral-03
- positive -> positive-01
- negative -> negative-01

**Color**
- primary -> neutral-01
- secondary -> neutral-02
- tertiary -> neutral-03
- positive -> positive-01
- negative -> negative-01

**Border**-color
- primary -> subtle-01
- secondary -> strong

----


astrit:
The Icon component will depricate few icons that should be removed with this release and also the sizing `width` and `height` property to be removed and communicate the change.

EldRoos:
Regression in small dropdown, the font is not smaller as it should.

astrit (approved):
🚀 let's goooooo 💚

PR #2250, opened 2025-06-25
Label: Reset styling to fix leaking

splashdust:
Re-open of accidentally closed #1886


splashdust:
@vsjolander Should we move this into the Chlorophyll 4 branch now?

PR #2263, opened 2025-07-01
Update context menu to use gds-button and menu-item fix

vsjolander:


astrit (changes requested):
@vsjolander 
We should probably use the label here as the button text and then icon as trail in this case

CleanShot 2025-07-01 at 21 23 01@2x

astrit (approved):

PR #2279, opened 2025-07-11
Fold out: Close when tab-key is pressed (#2)

HappyCtrl:
* Close fold out on `tab`

### DIFF
--- a/.changeset/new-hornets-play.md
+++ b/.changeset/new-hornets-play.md
@@ -1,0 +1,5 @@
+---
+'@sebgroup/green-angular': patch
+---
+
+Fold out: Close fold out when tab key is pressed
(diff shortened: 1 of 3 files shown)


EldRoos:
What happens if the component have interactable elements inside it? You can't tab to those (use keyboard to reach them)?

HappyCtrl:
> What happens if the component have interactable elements inside it? You can't tab to those (use keyboard to reach them)?

Yeah that's true, good point. My interpretation was that the intended use of this component is to display a list of elements, like a dropdown and then those should be reached with the keyboard arrows. But of course, if it's used in another way you're totally correct. I'll loop it back and think it over once more 😊

HappyCtrl:
> What happens if the component have interactable elements inside it? You can't tab to those (use keyboard to reach them)?

A little update 😊 The main reason for this approach was to have options with custom styling and event emitters etc from the parent consumer. And that the intended purpose of this component is/was not to have any navigational content inside it (thus nothing that should be "tabbable").

astrit:
@HappyCtrl @EldRoos ! 

Are we happy with this, is yes please approve to release with upcoming PR!

EldRoos:
Ok if it never have anything you can interact with other than keys than I guess this is ok.

astrit (approved):

Issue #1250, opened 2024-03-08
Datepicker: Allow setting timezone offset


splashdust:
By default, the Datepicker emits a `Date` object with the time set to 00:00 local time. In some cases it may be desirable to select dates dates in different time zone offset.

The API for this could look something like this:

```html
<!-- Explicitly select dates in CET time -->
<gds-datepicker
  utc-offset="+1"
></gds-datepicker>
```

~~I think it will still make sense to always default to setting the hours and minutes of the `Date` object to 00:00. If you need a different time you can always just add xx:xx to that date.~~

Update: Time component should be set to 12:00 by default.

Related to #1227

Issue #1227, opened 2024-02-23
Datepicker: Incorrect date when selecting older dates in reactive forms


sixskulls:
### Bug already reported?

- [X] I confirm that I have checked if the bug already has been reported

### For which framework/library you are reporting the bug

Angular

### Component name

datepicker

### Description

When using in reactive form datepicker value is incorrect when you select a previous date.

### Steps To Reproduce

1. In https://sebgroup.github.io/green/latest/angular/?path=/docs/components-datepicker--form
2. Select previous date in same month
3. See wrong date output in form value 
![image](https://github.com/sebgroup/green/assets/33746492/aef35392-d54b-4be3-8e9f-00c508f97b49)

### Current Behaviour

Form values returns previous day 22:00, instead of selected date. 

### Expected Behaviour

When selecting a specific date getting same date value in the form value.

splashdust:
@sixskulls The Datepicker emits a javascript `Date` object set to 00:00 of the selected date in local time. What you're seeing on the Storybook page is the ISO 8601 representation of that date in UTC offset (the Z at the end means UTC). So if you are on CET, for example, the ISO date will show -2 hours, which in this case will be 22:00 of the previous date.

It might make more sense for the Datepicker to set the time to 12:00 instead of 00:00, so that the date is the same in all time zones. I will look into it!

sixskulls:
Makes sense, I was wondering what I was missing. Feel free to close, we can adjust accordingly on our side.

splashdust:
Closing this one, but opened related #1250

PR #2294, opened 2025-07-23
Utc offset

EldRoos:
Closes: #1250
Related: #1227

Needs to be added to month-picker and year picker later on (other active pr).

### DIFF
--- a/libs/core/src/components/datepicker/datepicker.component.ts
+++ b/libs/core/src/components/datepicker/datepicker.component.ts
@@ -125,6 +125,12 @@ class Datepicker extends GdsFormControlElement<Date> {
   hideTodayButton = false
 
   /**
+   * Whether to override the default hour set of mid day (12:00) in utc. Eg. -2 sets time to utc 10:00.
+   */
+  @property({ type: Number, attribute: 'utc-offset' })
+  utcOffset = 0
+
+  /**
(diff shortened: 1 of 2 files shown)


splashdust (changes requested):

libs/core/src/components/datepicker/datepicker.component.ts
```diff
@@ -825,7 +834,7 @@ class Datepicker extends GdsFormControlElement<Date> {
     newDate.setFullYear(parseInt(this.#spinnerState.year))
     newDate.setMonth(parseInt(this.#spinnerState.month) - 1)
     newDate.setDate(parseInt(this.#spinnerState.day))
-    newDate.setHours(12, 0, 0, 0)
+    newDate.setHours(this.utcHours, 0, 0, 0)
```
splashdust, line 837: The function here needs to be changed to [setUTCHours()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setUTCHours), so that TZ is disregarded.

splashdust (changes requested):

libs/core/src/components/datepicker/datepicker.component.ts
EldRoos: I think I got it to work. There is no setUTCHours(), so I had to use DATE.setUTCHours() (on the other place).

splashdust (approved):

Issue #2293, opened 2025-07-23
Datepicker: Spinner click and mousewheel area


EldRoos:
### Does the component already exist

- [x] I confirm that I have checked if the component already exists in any of the Green libraries

### Description

Cool functionality to increase and decrease the datepicker input (spinners) with the mousewheel.
The font line-height is 18px, so it's a bit hard to keep the mouse exactly over where you want to scroll. WCAG recommends 48px for interactable elements. Probably hard to have it that size without breaking the layout, but should be able to have the scrolling area larger, at least a few pixels to the sides and 5-10 pixels over and under (line-height 32px works on the non small version).

Same with clicking at the different spinners, now you always select the first spinner even if you are much closer to the last but a bit outside it.

### Who will benefit from this feature?

Those that doesn't have super good mouses/low movement settings or have not the best hand movement control.

PR #2300, opened 2025-07-24
Datepicker: Bigger line height on spinners

EldRoos:
Closes: #2293

Might be better to use some const/figma variable instead of rem 2 and 1.5?


splashdust:
@EldRoos I think it might be better to change the flex alignment of `.main-slot-wrap` in the `gds-field-base` primitive to `stretch`. That way, the spinbuttons will take up the full height (minus padding) without any magic numbers.

A caveat is that it will affect other components slightly. Select and Dropdown needed adjustment when I tried it out in DevTools. Specifically, `align-self: stretch` needed to be added to the `.native-control` class in `shared-styles/form-control-host`

EldRoos:
Ah nice you got it to work :)

PR #2318, opened 2025-08-06
V-Angular - Dropdown - Added "close on scroll" function 

JohanAtSEB:
### nggv-dropdown
- Added input `closeOnScroll`, to be able to have the flexibility to toggle the function on/off
  - Default = `false`
- Created subscription `onScrollSubscription` and function `subscribeToOutsideScrollEvent`. Mirrors same functionality for `click`
  - If the dropdown is expanded and the scroll is happening **outside** of the component, that will say when the cursor is outside of the list, or when the list is too short to be scrollable even when the cursor is inside of the list, then the dropdown will close and subscription will be unsubscribed.
- If `closeOnScroll === true`, add class `close-on-scroll` to the component wrapper.
  - It's necessary to add `overscroll-behavior: none;` to the `<ul>`. Otherwise it will close the list if you are at the bottom of it and you still scroll down more. During my testing this happened a few times and I found it annoying and unexpected.
  -  A little detour was taken to add this to the `<ul>` in `dropdown-list` component. I didn't feel that it was the best way to send `closeOnScroll` boolean to the `dropdown-list` component just to add a css class, since all the logic of this is in `dropdown` component. It's against best practice, but I still believe this is the best option.
- Created new template `WithCloseOnScroll` for `dropdown` in storybook

### DIFF
--- a/.changeset/odd-dryers-jump.md
+++ b/.changeset/odd-dryers-jump.md
@@ -1,0 +1,5 @@
+---
+'@sebgroup/green-angular': patch
+---
+
+V-Angular: Add column label for Table component
--- a/libs/angular/src/v-angular/table/table.component.html
+++ b/libs/angular/src/v-angular/table/table.component.html
@@ -155,8 +155,8 @@
               </span>
               <span class="gds-item-field-text">
                {{ item[column.property] }} ({{ item.subItems.length }})
-                <div *ngIf="!!column.extraLabel" class="gds-item-field-extra-label">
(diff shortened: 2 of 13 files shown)


splashdust:
@JohanAtSEB Consider making the close-on-scroll behavior non-optional, unless there is a clear use-case for disabling it. Less configurability means more consistency!

Also, this behavior is not optional in the dropdown in Green Core, and aligning the features as much as possible is beneficial, since the idea is to merge these components at some point in the future.

PR #2320, opened 2025-08-07
V-Angular - Dropdown - Added "close on scroll" function

JohanAtSEB:
### nggv-dropdown
- Added functionality to close `dropdown-list` if you scroll outside of it
- Created subscription `onScrollSubscription` and function `subscribeToOutsideScrollEvent`. Mirrors same functionality for `click`
  - If the dropdown is expanded and the scroll is happening **outside** of the component, that will say when the cursor is outside of the list, or when the list is too short to be scrollable even when the cursor is inside of the list, then the dropdown will close and subscription will be unsubscribed.
- Added `overscroll-behavior: none;` to `ul[role='listbox']`. Otherwise it will close the list if you are at the bottom of it and you still scroll down more. During my testing this happened a few times and I found it annoying and unexpected.
- Created new template `CloseOnScroll` for `dropdown` in storybook, to be able to showcase the functionality

--

Continuation of https://github.com/seb-oss/green/pull/2318, as I accidentally deleted the branch

### DIFF
--- a/.changeset/odd-dryers-jump.md
+++ b/.changeset/odd-dryers-jump.md
@@ -1,0 +1,5 @@
+---
+'@sebgroup/green-angular': patch
+---
+
+V-Angular: Add column label for Table component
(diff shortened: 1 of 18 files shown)


splashdust (approved):
👌

PR #2341, opened 2025-08-20
Calendar & Datepicker: GDS23 style

EldRoos:
**Main changes:**

Updated Calendar to gds23 styling:
* Day names just one char.
* Added small size version.
* More spacing for indicator if needed.
* New icon type of indicator.
* Updated stories to reflect the changes.
* Made the styles general so they will work with month- and yearpicker too later on (styling will be moved to shared).

Updated Datepicker to gds23 styling:
* Primary today button.
* Fixed story text error of medium size, it's large.
* Fixed datepicker input font size, should be 16px when not small.


libs/core/src/components/calendar/calendar.component.ts
```diff
@@ -38,9 +38,10 @@ export type CustomizedDate = {
   color?: string
 
   /**
-   * The type of indicator.
+   * The type of indicator. Common types: `icon` (default)
+   * or `dot` (old default).
    */
-  indicator?: 'dot'
+  indicator?: 'icon'
```
vsjolander, line 44: This is a breaking change and should be added to the migration guide in the core-2 PR.

vsjolander (changes requested):
A work well done, good job!

Found some stuff that I commented on and also this:

Week number design looks a bit off. Can easily be confused as a date that is disabled?
Image

libs/core/src/components/calendar/calendar.component.ts
EldRoos: You are right about the week-numbers though, they should not be disabled in style.

splashdust (approved):
Everything looks goo to me now. Superb work @EldRoos! 💯

PR #2348, opened 2025-08-21
Button: Smaller font-size for small buttons

HappyCtrl:
Adjust font-size when for a button with class `small`, so font-size is set to 0.875 rem (14px).


splashdust (approved):

HappyCtrl:
Thanks, I totally forgot the component name in the changeset @splashdust

PR #2363, opened 2025-08-26
Swift package update

vsjolander:


libs/tokens/style-dictionary.build.ts
```diff
@@ -1,19 +1,44 @@
...
 import filters from './src/filters/index.ts'
 import formats from './src/formats/index.ts'
 import transforms from './src/transforms/index.ts'
 
-const swiftPackageName = 'GdsTokens'
+const swiftPackageName = 'GdsKit'
+const swiftTargets = [
+  {
+    name: swiftPackageName,
+    dependencies: ['GdsColours', 'GdsTypography', 'GdsDimensions'],
```
enerothseb, line 18: I think we're all used to using the American "Color" rather than the British "Colour"

libs/tokens/style-dictionary.build.ts
```diff
@@ -26,375 +51,539 @@ const capitalize = (str: string): string => {
...
+            destination: `figma.typography.json`,
+            filter: 'is-typography',
+          },
+          {
+            format: 'json/figma',
+            destination: `figma.size.json`,
+            filter: 'is-dimension',
+          },
+        ],
+      },
```
enerothseb, line 560: Why do we generate Figma inside ios theme?

Image

libs/tokens/src/templates/compose/color-scheme.template.ts
```diff
@@ -11,7 +11,7 @@ ${options.accessControl ? `${options.accessControl} ` : ''}${
 ${allTokens
   .map(
     (token) =>
-      `  ${token.comment ? `/** ${token.comment} */\n  ` : ''}${options.accessControl ? `${options.accessControl} ` : ''} ${formatProperty({ ...token, name: token.name.replace('sysColor', '') })}`,
+      `  ${token.comment ? `/** ${token.comment} */\n  ` : ''}${options.accessControl ? `${options.accessControl} ` : ''} ${formatProperty({ ...token, name: token.name })}`,
```
enerothseb, line 14: formatProperty doesn't seem to work the same way. `sysColor` is now included in the name for DarkModeColors and LightModeColors, but not in GdsColorTokens.kt, as you kept the string replacement in `data-class-template.ts`

PR #2369, opened 2025-08-27
Fix compose formatting

enerothseb:
In the 2.0 update, the string replacement for removing "sysColor" from the name was removed for the color-scheme-template, whereas it was kept for data-class-template. This caused the color names to be different in the two files, which does not compile on the Android side.


enerothseb:
> Maybe we should add a changeset for tokens so we trigger a new release? @enerothseb
> 
> I will break these out to their own packages aswell in a good way..

Okay @vsjolander, I've added a changeset.

PR #2383, opened 2025-08-29
Pagination+Alert: Color fix

EldRoos:
Was black color on black background for the selected tab in pagination.
Bug reported in GDS->General.

Color base-000 is not used and becomes black.

Edit: Same error in Alert success. Should be white text on green, not black text on green. Checked the old design.

### DIFF
--- a/libs/chlorophyll/scss/components/alert-ribbon/_mixins.scss
+++ b/libs/chlorophyll/scss/components/alert-ribbon/_mixins.scss
@@ -15,7 +15,7 @@ $variants: (
     ghost-button-type: 'light',
   ),
   'success': (
-    color: var(--gds-sys-color-base-000),
+    color: var(--gds-sys-color-white),
     border-color: transparent,
     link-outline-color:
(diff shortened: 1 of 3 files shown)


libs/chlorophyll/scss/components/alert-ribbon/_mixins.scss
```diff
@@ -15,7 +15,7 @@ $variants: (
     ghost-button-type: 'light',
   ),
   'success': (
-    color: var(--gds-sys-color-base-000),
+    color: var(--gds-sys-color-white),
```
splashdust, line 18: I think `--gds-sys-color-text-inverted` would be more appropriate here, otherwise it won't invert in dark mode.

Then again, not sure if this component even works in dark mode :)

EldRoos: Ok done. Background gets grey in dark-mode and text black, so I guess it's ok.

splashdust (approved):
👌

PR #2401, opened 2025-09-04
Add errorMessage property

nerjok:


astrit:
@nerjok just checking if this is still active!

astrit:
Closing for now this PR, potentially they will switch to using core components.
If the issue comes up again we can reopen this.

PR #2407, opened 2025-09-04
Adding manual dispatch of rc workflow

vsjolander:


splashdust (approved):
Looks good to me. Let's try this out and see if it is sufficient for what we need!

vsjolander:
Running it on the chlorophyll-4 PR now 🙏

PR #2408, opened 2025-09-05
Docs · Font, Tokens and Studio revamp

astrit:


astrit:
> Updated Studio is goosebumps-level good! Incredible work @astrit!
> 
> Only things I'd like to address before publishing is:
> 
> * Pointer cursor on cards on landing page
> * Set `tabindex="0"` on the cards to make them focusable (both landing and icons)
> * Set `role="link"` on the cards (both landing and icons)
> 
> Alternatively wrap the cards in anchor elements, but I guess that will cause a cascade of styling issues.

Thank you very much for the feedback ☘️ 🙏 

All three added and some interim outline and hover background as well. 
CleanShot 2025-09-26 at 08 48 09@2x

PR #2411, opened 2025-09-09
V-Angular Dropdown: move dropdown list to CDK portal

Jaroslavas-1:
**Overlay-based Dropdown List Refactor**
This PR refactors the dropdown component to use **Angular CDK Overlay** for rendering the dropdown list, improving positioning, responsiveness, and robustness. Key changes include:

- Dropdown list is now rendered in a CDK overlay (portal), allowing it to appear above other content and avoid clipping by scrollable containers.
- Overlay positioning uses FlexibleConnectedPositionStrategy with both "below" and "above" options, a configurable gap, and .withPush(false) to prevent overlap with the trigger.
- Overlay width is dynamically set to match the trigger button, and updates on window resize for responsiveness.
- Dropdown list height is calculated based on available viewport space (above or below the trigger), ensuring it fits the screen and avoids overflow.
- Scroll strategy is configurable via an input **closeDropdownListOnScroll**, allowing the dropdown to either close or reposition on scroll.

### DIFF
--- a/libs/angular/src/v-angular/dropdown/dropdown-list/dropdown-list.component.ts
+++ b/libs/angular/src/v-angular/dropdown/dropdown-list/dropdown-list.component.ts
@@ -79,12 +79,6 @@ export class NggvDropdownListComponent implements OnInit, OnChanges {
    */
   @Input() selectWithSpace = true
 
-  /**
-   * When true, the dropdown will automatically choose to open above or below the input
-   * based on available space in the viewport, and will scale its height to fit if needed.
-   */
-  @Input() dynamicPosition = false
-
   @Output() selectedValueChanged = new EventEmitter<any>()
 
   @Output() closed = new EventEmitter<void>()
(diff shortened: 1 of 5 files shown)


libs/chlorophyll/scss/index.scss
```diff
@@ -1,2 +1,5 @@
 @forward 'tokens';
 @use 'components';
+
+// used for v-angular dropdown
+@import '@angular/cdk/overlay-prebuilt.css';
```
splashdust, line 5: This is not ideal, since it forces everyone to include all of Chlorophyll to get this import. And many of the V-Angular components already bundles styles from Chlorophyll, so you end up with a lot of duplicated CSS if you do this (probably over 500kb of extra CSS)

Ideally, Angular would have a way to declare global style dependencies for a component, but AFAIK, there's no way to do that. So I think consumers will just have to add this import line manually. Maybe you can add a note about this in the dropdown stories?

libs/angular/src/v-angular/dropdown/dropdown-list/dropdown-list.component.ts
splashdust, line 1: I see a few of what looks like breaking changes in here (i.e removed/changed inputs). Is it possible to keep the old ones around and mark them as deprecated?

splashdust (changes requested):

PR #2415, opened 2025-09-11
Card · Pattern 01

astrit:
Closes: 

- https://github.com/seb-oss/core-digital-design/issues/523
- https://github.com/seb-oss/core-digital-design/issues/556
- https://github.com/seb-oss/core-digital-design/issues/496
- https://github.com/seb-oss/core-digital-design/issues/478

### DIFF
--- a/libs/core/src/utils/mixins/props-image.ts
+++ b/libs/core/src/utils/mixins/props-image.ts
@@ -6,37 +6,11 @@ import { GdsElement } from '../../gds-element'
 type Constructor<T = GdsElement> = new (...args: any[]) => T
 
 export interface ImageProps {
-  /**
-   * The URL of the image
-   */
   src?: string
-
-  /**
-   * The srcset attribute for responsive images
-   * Format: "image.jpg 1x, image-2x.jpg 2x" or "image-320.jpg 320w, image-640.jpg 640w"
-   */
   srcset?: string
-
-  /**
-   * The sizes attribute for responsive images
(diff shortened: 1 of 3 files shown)


libs/core/src/utils/mixins/props-image.ts
```diff
@@ -0,0 +1,72 @@
...
+export interface ImageProps {
+  /**
+   * The URL of the image
+   */
+  src?: string
+
+  /**
+   * The srcset attribute for responsive images
+   * Format: "image.jpg 1x, image-2x.jpg 2x" or "image-320.jpg 320w, image-640.jpg 640w"
+   */
```
splashdust: The doc comments needs to go in the implementation for CEM analyzer to pick them up.

libs/core/src/utils/mixins/props-link.ts
```diff
@@ -0,0 +1,92 @@
...
+    /**
+     * URL that the link points to
+     * Can be absolute, relative, or fragment identifier
+     */
+    @property()
+    href?: string
+
+    /**
+     * Specifies where to open the linked document
+     * @property {'_self' | '_blank' | '_parent' | '_top'}
```
splashdust: No need to specify type information in doc comments, since that's already provided by Typescript

splashdust (changes requested):
Fantastic work as usual!

Only have a few nit-picks, mostly just some minor code convention stuff I noticed.

PR #2433, opened 2025-09-21
Improving Alert docs

vsjolander:


libs/core/src/components/alert/alert.stories.ts
```diff
@@ -23,29 +25,42 @@ const meta: Meta<GdsAlert> = {
...
+    },
   },
+  args: {
+    variant: 'information',
+    role: 'alert',
+    buttonLabel: '',
+    innerHTML: html`<strong>Information</strong> Body text starts on the same
+      row as heading. A link (optional) always ends the message.`,
+  },
+  render: (args) => html`
```
splashdust, line 41: It's actually better to leave out the render function, since it breaks Storybooks ability to automatatically assign args. With a render function, everything needs to be manually forwarded and it's easy to forget to update it later.

astrit: Updated this to use `argTablePropsFor` function and also no render except for variants story that has multiple `alert`

PR #2445, opened 2025-09-29
Breadcrumbs · Dynamic and restructuring

astrit:
Closes: https://github.com/seb-oss/core-digital-design/issues/576


libs/core/src/components/breadcrumbs/breadcrumb/breadcrumb.component.ts
```diff
@@ -0,0 +1,122 @@
...
+   * Accessible label for the breadcrumb item. Use when visual text needs a different description.
+   */
+  @property()
+  label = ''
+
+  /**
+   * Causes the browser to treat the linked breadcrumb URL as a download. Can be used with or without a value.
+   */
+  @property()
+  download?: string
```
splashdust, line 60: Do we need download for breadcrumbs? Maybe we can remove this and see if an actual need comes up later.

splashdust (commented):
Great work! Only found some very minor things to check before we merge, see the comments.

PR #2455, opened 2025-10-02
Filter-chips: enable-deselecting-single-chips

kobraan:
While in single select mode, users should be able to deselect a previously selected filter chip


splashdust (approved):
👌

Issue #826, opened 2023-04-03
Table component


VaidasMasys:
### Does the component already exist

- [X] I confirm that I have checked if the component already exists in any of the Green libraries

### For which framework/library you are requesting the component

Angular

### Description

It should be a basic reusable table component that allows a user to sort data on columns. 

### Who will benefit from this feature?

If this table is done right in the long term pretty much every team that uses tables can benefit from this. In payments I believe all teams uses tables and they use a variety of different tables. They often look a bit different and sometime work a bit different. One table would help to unify designs and functionalities of tables within SEB Angular projects.

PR #2494, opened 2025-10-27
Table

astrit:
Closes:

- [x] https://github.com/seb-oss/core-digital-design/issues/624
- [x] https://github.com/seb-oss/core-digital-design/issues/625
- [x] https://github.com/seb-oss/core-digital-design/issues/626
- [x] https://github.com/seb-oss/core-digital-design/issues/627
- [x] https://github.com/seb-oss/core-digital-design/issues/628
- [x] https://github.com/seb-oss/core-digital-design/issues/629
- [x] https://github.com/seb-oss/core-digital-design/issues/630

Related: #826

### DIFF
--- a/apps/docs/design/layout/home/home.tsx
+++ b/apps/docs/design/layout/home/home.tsx
@@ -168,7 +168,7 @@ export function Home({
               href={`/component/${component.slug}`}
               key={component.slug}
             >
-              <GdsFlex flex-direction="column" height="240px">
+              <GdsFlex flex-direction="column" height="280px">
                 <GdsCard
                   width="100%"
                   padding="m"


libs/core/src/components/table/table.component.ts
```diff
@@ -0,0 +1,1285 @@
...
+   * > Variant naming will change in the next major release.
+   */
+  @property()
+  variant: 'primary' | 'secondary' | 'tertiary' = 'secondary'
+
+  @state()
+  private _isMobile = false
+
+  @watchMediaQuery('(max-width: 768px)')
+  _handleMobile(matches: boolean) {
```
splashdust: Missing `private` keyword

libs/core/src/generated/locales/da.ts
splashdust: Looks like all the languages are still english strings!

.changeset/tender-hornets-reflect.md
splashdust: Can't see this change in the PR. Was it reverted?

splashdust (changes requested):
Superb work on both the table and the pagination component. This PR is a true masterpiece!

Only found a few bits and bobs that I think would be good to fix before merging.

PR #2507, opened 2025-11-05
Button: Ripple effect now triggers on keyboard press (Space/Enter)

zygispr:
Now the ripple effect triggers if focus is on Button and user presses Space or Enter. Previously it would only trigger on mouse click


splashdust (approved):
Looks great, thank you for this! 💚

PR #2509, opened 2025-11-06
Dropdown: gds-blur event is fired only when focus is shifted away from Dropdown component

zygispr:
There was an unwanted behavior with the Dropdown component, where the `blur` event would fire twice when interacting with the component:
- first when you open the dropdown and the focus is manually shifted to the first option of the expanded dropdown
- secondly when the dropdown closes and you shift focus away from it. 

Normally you'd expect the `blur` to trigger once after dropdown is closed and focus is shifted away from it.

Fixing the normal behavior of `blur` didn't seem possible, but we can fix it by listening to the `blur` event and only firing a custom `gds-blur` event when it's needed.

### DIFF
--- a/.changeset/sweet-foxes-rush.md
+++ b/.changeset/sweet-foxes-rush.md
@@ -1,5 +1,0 @@
----
-'@sebgroup/green-core': minor
----
-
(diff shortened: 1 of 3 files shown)


splashdust (approved):
Perfect! 👌

PR #2522, opened 2025-11-18
chore(core): export tokens.style.js

patrikfogelstrand:
Normally, I import any Green Core component (and potentially using GdsTheme if needed) and the CSS variables are loaded through those imported components. 

However I have found that when mixing routing engines (such as doing a full reload vs. doing a client side navigation using e.g. React Router), the tokens are sometimes lost. Could be a race condition or how I am configuring the client (in this case [Docusaurus](https://docusaurus.io/)), but I think it would be solved by always calling the `tokens.style.js` file in a [Docusaurus client module](https://docusaurus.io/docs/advanced/client#client-modules).

Perhaps there could be other use cases for this also? 

The encouraged way to import the tokens should be directly from importing Green Core components or using GdsTheme.

### DIFF
--- a/.changeset/moody-dancers-flow.md
+++ b/.changeset/moody-dancers-flow.md
@@ -1,0 +1,5 @@
+---
+'@sebgroup/green-core': patch
+---
+
+chore: export tokens.style.js for consumers to load css variables using js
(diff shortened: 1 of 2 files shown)


patrikfogelstrand:
Having some pipeline issues, which blocks deployment when testing alternative solutions. Will get back to this asap.

PR #2526, opened 2025-11-25
Segmented control · Fix selection color

astrit:
Closes: 

- https://github.com/seb-oss/core-digital-design/issues/653


splashdust (approved):
👌

PR #2579, opened 2026-01-21
Table · Events

astrit:
## Table pagination and sorting events

Add `gds-page-change`, `gds-rows-change`, and `gds-sort-change` events to Table so external listeners can react to pagination and sorting changes.

### DIFF
--- a/.changeset/bold-results-show.md
+++ b/.changeset/bold-results-show.md
@@ -1,0 +1,5 @@
+---
+'@sebgroup/green-core': minor
+---
+
+**MCP:** Add experimental local MCP server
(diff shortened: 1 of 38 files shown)


libs/core/src/components/table/table.component.ts
```diff
@@ -1160,16 +1163,30 @@ export class GdsTable<T extends Types.Row = Types.Row> extends GdsElement {
...
-        this._view.sortColumn === columnKey &&
-        this._view.sortDirection === 'asc'
-          ? 'desc'
-          : 'asc',
+      sortDirection: newDirection,
       page: 1,
     }
+
+    // Dispatch event for external listeners
+    this.dispatchEvent(
```
splashdust: For the event dispatch, we need to use `this.dispatchCustomEvent` from GdsElement, so that camelCased versions are dispatched as well.

splashdust (changes requested):

PR #2593, opened 2026-01-29
Table · Bugfixes

astrit:


libs/core/src/components/link/link.component.ts
```diff
@@ -98,4 +99,17 @@ export class GdsLink extends withMarginProps(
...
+   * listeners can intercept them and call preventDefault() for custom routing logic.
+   * Without this, clicks would be trapped in the shadow DOM.
+   */
+  #handleClick(e: MouseEvent) {
+    this.dispatchCustomEvent('gds-link-click', {
+      bubbles: true,
+      composed: true,
+      detail: e,
+    })
+  }
```
splashdust: Click events are composed by default, so this should not be necessary. To find out which gds-link element was clicked in a table, a consumer can just follow the `composedPath` of the event, for example:

```ts
elTable.addEventListener('click',
  (event) => {
      const linkElement = e.composedPath().find(el => el.gdsElementName === 'gds-link')
      console.log('Clicked link: ', linkElement)
   }
)
```

It's a bit cumbersome, but the procedure would the same with the custom event as well, just a different event name.

To make it easier for the consumer, I think it would be better to add helper functionality in table instead.

